PR #15, opened 2026-06-17
feat(connect): ship a no-database sample (agami-example) + Phase 0s onboarding

ashwin-agami:
## What this does

Lets a user try agami with **zero connection setup** — a small, deterministic SQLite sample (retail + subscriptions, "Acme Store") plus a prebuilt, validated semantic model. From install → first answered question in **under a minute**, with nothing leaving the machine.

Entry: pick **"Try a sample database"** in the connect picker, or run `/agami-connect sample` / say "I don't have a database". Creates the **`agami-example`** profile, untouching any existing profiles.

## What's in it

**Sample (committed text only, ~160K — the built `.db` is gitignored):**
- `samples/store/seed.sql` — deterministic generator (no `random()`), ~22k rows → ~1.5MB. Exercises a **fan trap**, a **chasm trap**, enum breakdowns, PII columns, and ~24 months of dated rows for trends.
- `samples/store/build_sample.py` — stdlib builder (`sqlite3` CLI + `executescript` fallback). Byte-identical across both methods and across runs.
- `samples/store/model/` — prebuilt model for `agami-example`, authored via the **real introspect → enrich → seed pipeline**: described tables, enum labels, USD units, PII flags, 8 entities, 8 signed-off metrics, **14 EXPLAIN-validated seeds** (incl. two multi-table-join showcases). Synthetic data only; generic `agami-team` sign-off identity.

**agami-connect:**
- "Try a sample database" picker option **and** a preflight step-0 short-circuit so the sample is reachable even when another profile is active.
- New **Phase 0s**: build → write creds (via the existing promoter) → `.config` → **6A copy / 6B live-rebuild** fork → describe + offer questions, then **stop** (no scripted demo). "Where to go deeper" (`/agami-model`, `/agami-serve`, corrections) moves to **after** the first answer.
- Skill-wide **plain-voice** rule (no "wow moment"/marketing voice in todos, progress, or narration).

## Bug fixes surfaced while building this (general, not sample-specific)
- **`execute_sql`**: add missing `import json` — every fan/chasm pre-flight refusal in the Python tier was crashing with `NameError` instead of returning the refusal message.
- **`loader`/`cli`**: include `performance_hints` in the default context/bundle assembly so `estimated_row_count` reaches the answer receipt (fixes the **"rows unknown"** provenance) and the scan-risk check.
- **agami-query**: stop narrating the query-log write / `.optins` gate check.

## Tests
`tests/test_sample_database.py` (18) — determinism, row counts, model validation, fan/chasm refusal, row-count-in-context guard, and an `execute_sql` refusal-path regression guard. Full unit suite: **589 passing**.

## Consistency note
The model's auto-derived names (`organization`, area, `agami-example_sqlite` connection) all derive from the profile, so the **6A copy** and **6B live rebuild** produce the same structure — verified by re-introspecting under the profile name rather than hand-editing.

## Deliberately out of scope (tracked follow-ups)
- **Mechanical render piping** — `render_chart.py` should source displayed values from the result CSV instead of the LLM transcribing them (real correctness/trust gap; touches every render).
- **Snapshot-writer** — `model_version` is `null` for *every* profile because nothing writes `.snapshots/`; needs an engine-side writer at introspect/curate.
- **Churn metrics** in the sample (optional polish).
- Per-query PII opt-in override — considered, **declined** for now (masked-by-default stands).


ashwin-agami:

PR #16, opened 2026-06-17
feat(model): write model-version snapshots (fix model_version: null everywhere)

ashwin-agami:
## Problem

`<profile>/.snapshots/<hash>/` is read by `mcp_server._model_version` and the agami-query skill (newest dir name = the model version reported in every answer's trust receipt), and it's documented in `file-layout.md` and the chart template — but **nothing ever wrote it**. So `model_version` was `null` for **every** profile, on every answer. The trust-receipt's version pin was silently inert.

## Fix

New `semantic_model/snapshot.py`:
- **`compute_model_hash(root)`** — 12-char SHA-256 over the model's content: every file under the profile root *except* machine-state (`.snapshots/`, `.introspect/`, `.legacy_backup/`, `.git/`, `curation_log.jsonl`). Hash changes iff the model changes.
- **`write_snapshot(root)`** — creates `.snapshots/<hash>/` with a **deterministic** `manifest.json` (model hash + per-file hashes; no timestamp, so a committed snapshot never churns — recency comes from the dir mtime, which is exactly what the readers use). **Idempotent** (an unchanged model re-uses its dir, mtime-refreshed so it stays "newest"); **prunes** to the last 20. Best-effort — never raises into the write path.

Stamped at the **two deterministic model-write chokepoints**, so the skill, the local MCP server, and cron all get a version identically:
- `build.write_tree` (introspect)
- `curate._git_commit` (every curation / enrichment / example write) — placed **before** the git check so it runs whether or not the artifacts dir is a git repo (it usually isn't).

## Tests

`tests/test_model_snapshots.py` — hash stability + change-sensitivity, machine-state exclusion, idempotent dir, prune cap, the `mcp_server._model_version` reader contract (newest-by-mtime wins after a change), and the **real introspect path** stamping a snapshot. **578 unit tests pass.**

## Scope note

This makes `model_version` **resolve**. Full content-reproduction "against the hash it ran on" (storing the model state to roll back to) remains a future enhancement — the manifest records per-file hashes for *verification*, not the content to reconstruct.

### DIFF
--- a/plugins/agami/scripts/semantic_model/build.py
+++ b/plugins/agami/scripts/semantic_model/build.py
@@ -534,6 +534,10 @@ def write_tree(
         if sa.name in examples_by_area:
             write(f"prompt_examples/{sa.name}/examples.yaml",
                   _dump({"examples": examples_by_area[sa.name]}))
+    if not dry_run:
+        # Stamp the model_version pin for the freshly-written model (best-effort).
+        from . import snapshot
+        snapshot.write_snapshot(out)
     return rep
 
 
--- a/plugins/agami/scripts/semantic_model/cli.py
+++ b/plugins/agami/scripts/semantic_model/cli.py
@@ -56,6 +56,17 @@ def cmd_validate(args) -> int:
     return 0 if res.ok else 1
 
 
+def cmd_snapshot(args) -> int:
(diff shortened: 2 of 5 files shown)


ashwin-agami:

PR #17, opened 2026-06-18
feat(query): deterministic render piping — numbers from the CSV, not the LLM (Phase A)

ashwin-agami:
First phase of the determinism refactor (anchor doc: `docs/design/determinism-refactor.md`).

## Problem
The render path had the LLM **hand-build `agami-sections.json`, transcribing result numbers** into both `table_rows` and `datasets[].data`, and hand-build the trust receipt. `render_chart.py` never reads the result CSV — so a miscopy shows a wrong number in the *chart* while the *table* is right, and nothing catches it. For a tool whose pitch is "trust these numbers," that's the wrong place to rely on the model re-typing.

## Fix — take numbers + provenance out of the model's hands
- **`scripts/csv_to_sections.py`** — reads each section's result CSV + the `units` map (from `sm prepare`) and emits `labels` / `table_rows` (formatted *exactly* via the same `units.py` `sm format-table` uses) / `datasets[].data` (raw numbers). The LLM supplies only a **presentation spec**: title, insight, chart_type, which column is the label vs the values, cosmetic relabels, and the `sql_file` (read verbatim — no paraphrase). Surfaces **anomalies** (a non-numeric value column, 0 rows) for the LLM rather than guessing.
- **`sm receipt`** (`cli.cmd_receipt`) — wraps the existing **`runtime.assemble_receipt`** (the *same* builder the MCP server already uses), so the receipt is parsed from the SQL + model (tables, relationships, metrics, warnings, assumptions, `model_version`) instead of hand-extracted. The LLM only appends an *ad-hoc* metric it composed.
- **agami-query Phase 4e.iii / 4e.iii.5** rewritten to call them; the old "transcribe the formatted values… `datasets[].data` is the only place raw numbers belong" prose is removed.

## Contract (settled here for Phases B/C)
Refactor scripts print `{ok, data, anomalies[], needs_judgment?}`. Anomalies are non-fatal signals; `needs_judgment` is the escape hatch where a script deliberately stops short of guessing and hands context to the LLM (e.g. the "samples data, finds an anomaly, figures out what to do" case).

## Tests
`tests/test_render_piping.py` — number fidelity (table formatted `$121,560.94`, datasets raw `121560.94`, exact big values), non-numeric-col anomaly (not a crash), SQL read verbatim from file, mixed-unit section (no false chart unit), multi-section; `sm receipt` from SQL (tables + relationships parsed). **579 unit tests pass.**

## Scope
Phase A only (render). Phase B (connect bootstrap spine) and Phase C (collapse near-deterministic skills) are separate PRs per the design doc. Independent of the open sample (#15) and snapshots (#16) PRs.

### DIFF
--- a/docs/design/determinism-refactor.md
+++ b/docs/design/determinism-refactor.md
@@ -1,0 +1,127 @@
+# Determinism refactor — move mechanical orchestration out of LLM prose
+
+**Status:** in progress · **Owner:** agami core · anchor doc for a multi-PR effort.
+
+## Why
+
+agami's skills are LLM prose that interleaves *judgment* (what does the user
+mean? what does this column mean? is this number real?) with *mechanical
+plumbing* (resolve a path, parse a pasted block, format a number, write a
+config). When the LLM has to faithfully execute the plumbing, it sometimes
+doesn't — the fidelity failures we keep hitting are almost all in the plumbing,
+not the judgment: a wrong interpreter written to `.config`, a `/sample` vs
+`/agami-example` mis-wire, a `sqlite3` query missing `-header`, the model
+**re-typing result numbers** into a chart (a silent wrong-number risk).
+
+The deterministic *core* is already heavily scripted (13 scripts + ~30 `sm`
+subcommands). The gap is the **orchestration** around it, still in prose.
+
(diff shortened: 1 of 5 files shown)


ashwin-agami:

PR #37, opened 2026-06-25
ACE-001: HTTP MCP transport + auth shim (factor TOOLS; smart get_datasource_schema)

ashwin-agami:
Spec: ACE-001

## Summary

Turn the stdio-only MCP harness into a **shareable HTTP MCP server** without forking the tools: factor the `TOOLS` registry so the stdio and HTTP entrypoints share one implementation, reduce both surfaces to the **same 5 product tools**, build the adaptive `get_datasource_schema`, and add the HTTP transport + `401`/`WWW-Authenticate` auth shim + `.well-known` OAuth discovery.

## Changes

- **`tools.py`** — the shared registry + the 5 product-tool impls + helpers + server identity. `mcp_harness.py` is now a thin stdio JSON-RPC adapter that imports it.
- **5-tool surface on both transports** — `list_datasources`, `get_datasource_schema`, `get_prompt_examples`, `execute_sql`, `log_feedback`. The granular traversal tools, `identify_entity`, `pre_flight_check`, and `save_correction` are dropped from MCP (their capability stays on the CLI/skill).
- **Adaptive `get_datasource_schema`** — `mode=auto` picks full/summary/index by **subject-area count** (≤12 / ≤50 / 51+); a ~60K-char budget downgrades one rung at a time even for a forced `full` (`truncated=true`); `query`/`metric_names` lexically rank metrics (exact/substring always kept, weak tail ≥0.6 coverage capped top-10); `metric_index` + `large_tables` always present. `dataset_names` = explicit full scope, no downgrade.
- **`mcp_http.py`** — a streamable-HTTP MCP transport (official SDK low-level `Server`) whose surface IS `tools.TOOLS` (no duplicate defs); a bearer-presence auth shim (unauth → `401` + `WWW-Authenticate resource_metadata`); the `.well-known/oauth-protected-resource` + `…/oauth-authorization-server` discovery endpoints advertising `PUBLIC_BASE_URL`. Behind a new **`[server]`** extra (MCP SDK + ASGI).
- `execute_sql` (the executor chokepoint) is **unchanged** — moved verbatim, single shared impl both transports route through.

## Decisions (PR body, per the rubric)

- **HTTP isolated in `mcp_http` + `[server]` extra** so the local executor/skill stay stdlib-lean; the no-network contract test is expanded to cover the relocated library and **excludes `mcp_http`** as the one deliberate network surface.
- **5-tool reduction changes the stdio surface (11→5)** intentionally — Claude Desktop now matches claude.ai; dropped capabilities remain on the CLI.
- **`PUBLIC_BASE_URL` is required** (no proxy auto-detection) — it backs discovery + the `WWW-Authenticate` resource URL.
- Auth is **presence-only** (requires the Bearer scheme); real token validation + the authorize page are a later feature.
- Smart-schema output extends the local schema contract (forward-compatible).

## Verification

- `uv run dev.py check` green: ruff · **686 tests** · gitleaks. CI installs `[model,server]` so the HTTP tests run.
- **Both transports return the identical 5** — stdio (subprocess JSON-RPC) and HTTP (authed `initialize`→`tools/list` via TestClient); they share one registry object (asserted).
- Adaptive sizing proven on a synthetic model (auto tiers, 60K-budget downgrade + `truncated`, query metric ranking, `large_tables`).
- Unauthenticated / non-Bearer / empty-token HTTP requests → `401` + `WWW-Authenticate`; discovery advertises `PUBLIC_BASE_URL`.
- `execute_sql.py` unchanged vs main.

## Checklist

- [x] Spec-linked; in scope (no authorize page / Postgres / config hardening)
- [x] Ports-and-adapters: shared core + thin entrypoints; core imports neither transport
- [x] Both surfaces in sync, tested on both
- [x] New behavior has tests; no tests weakened (save_correction tests removed with the tool)
- [x] Full local gate green
- [x] No real customer data/names/credentials; the only network surface is the explicit HTTP transport
- [x] No internal spec/roadmap IDs in repo files (only this `Spec:` line)

## Out of scope

OAuth authorize page / login / identity; Postgres-backed serving; cloud-neutral config hardening — each a separate spec.


packages/agami-core/src/mcp_harness.py
```diff
@@ -56,894 +29,23 @@
...
-CONFIG_PATH = agami_paths.config_path()
-QUERY_LOG = agami_paths.query_log_path()
-FEEDBACK_LOG = AGAMI_LOCAL / "feedback.jsonl"
-
-SERVER_NAME = "agami"
-# The protocol version we fall back to if the client doesn't pin one. We echo
-# the client's requested version when present (see _handle_initialize).
+# The protocol version we fall back to if the client doesn't pin one. We echo the client's
+# requested version when present (see _handle_initialize).
 DEFAULT_PROTOCOL_VERSION = "2024-11-05"
```
ashwin-agami, line 47: I dont quite understand this

packages/agami-core/src/tools.py
```diff
@@ -0,0 +1,653 @@
...
+    "never hide them. Don't refuse on an unreviewed metric — answer and warn.\n"
+    "PII: a column marked `sensitive: true` restricts OUTPUT, not the query — you MAY "
+    "COUNT/COUNT(DISTINCT)/filter/GROUP BY/JOIN on it, but never SELECT its raw per-row values. "
+    "'unique emails' → COUNT(DISTINCT email). To disambiguate identical labels, project the "
+    "non-sensitive id. (execute_sql enforces this and errors on a raw sensitive projection.)"
+)
+
+
+def bootstrap_paths() -> None:
+    """Run the one-shot ~/.agami migration, then re-resolve the module-level paths (the migration
```
ashwin-agami: We dont do any ~/.agami folders anymore right?

packages/agami-core/src/tools.py
```diff
@@ -0,0 +1,653 @@
...
+    """Local analog of Ask Agami `get_prompt_examples`: the few-shot library.
+
+    Returns the curated examples.yaml verbatim as text. Local serving has no
+    embedding store, so `query`/`top_k` are accepted for interface-parity with
+    the hosted connector but do not rank or cap — the full curated library is
+    returned (it is small; the client reads YAML directly).
+    """
+    profile = resolve_profile(args.get("datasource"))
+    artifacts = resolve_artifacts_dir()
+    ex_dir = artifacts / profile / "prompt_examples"
```
ashwin-agami, line 640: So these are all local file based. I am assuming we will change this for the future PRs when we package the MCP server?

packages/agami-core/src/tools.py
```diff
@@ -142,9 +144,17 @@ def _db_type_for(profile: str, creds: dict[str, dict[str, str]]) -> str:
     t = sect.get("type", "")
     if not t and sect.get("url"):
         scheme = sect["url"].split("://", 1)[0].split("+", 1)[0].lower()
-        t = {"postgresql": "postgres", "postgres": "postgres", "mysql": "mysql",
-              "mariadb": "mysql", "redshift": "redshift", "snowflake": "snowflake",
-              "bigquery": "bigquery", "bq": "bigquery", "sqlite": "sqlite"}.get(scheme, scheme)
+        t = {
```
ashwin-agami, line 152: For now, are these the only DBs we support? No SQL Server or ORacle and such? Its okay if we don't support them, but might be worth adding to DB for now.

packages/agami-core/src/tools.py
```diff
@@ -294,13 +315,222 @@ def _distill_for_llm(text: str | None) -> str:
...
+_AUTO_SUMMARY_MAX_AREAS = 50  # <= this -> summary; above -> index
+_SCHEMA_CHAR_BUDGET = 60_000  # decoded len(json.dumps(...)) ceiling (~15K tokens)
+_SCHEMA_MODE_DOWNGRADE = {"full": "summary", "summary": "index", "index": None}
+_LARGE_TABLE_ROWS = 1_000_000  # tables at/above this surface in `large_tables` in every mode
+
+# Metric ranking (lexical, no embeddings): exact/substring hits ("strong") are always kept; the
+# weak token-overlap tail needs >= this coverage and is capped at top-K.
+_METRIC_MATCH_TOP_K = 10
+_METRIC_MATCH_MIN_COVERAGE = 0.6
+_METRIC_MATCH_STOPWORDS = frozenset(
```
ashwin-agami, line 349: what if words fall outside this list? would the LLM still detect it? What's the core use for this?

packages/agami-core/src/mcp_harness.py
ashwin-agami: MCP negotiates a protocol version in the \`initialize\` handshake: the client sends the version it wants and the server echoes it back to accept it. This constant is only the fallback we assume when a client connects *without* naming one (older clients) — not a pin. Reworded the comment to say that in 0706488.

packages/agami-core/src/tools.py
ashwin-agami: Correct — new installs use \`<artifacts_dir>/local\` only; nothing writes \`~/.agami\` anymore. \`agami_paths.bootstrap()\` still runs a one-time migration of any *legacy* \`~/.agami\` install into the new layout, so an older user upgrading doesn't lose their setup; it's a no-op once there's nothing to move. It's a backward-compat shim we can drop in a later cleanup. Clarified the docstring in 0706488.

packages/agami-core/src/tools.py
ashwin-agami: Yes — exactly. Everything model/log-related here is local file/YAML by design (the free local product). ACE-002 (Postgres-backed serving) swaps the persistence behind the same tool impls via the OCR-029 ActivitySink/loader ports — the tool code here doesn't change, only what it reads/writes. That's the next spec in this feature, after this lands.

packages/agami-core/src/tools.py
ashwin-agami: Good catch — that map was narrower than what we advertise. Extended it in 0706488 to cover SQL Server (mssql/sqlserver), Oracle, Databricks, Trino/Presto, and DuckDB; an unknown scheme passes through verbatim. Note this only sets the **display** \`database_type\` label in list_datasources — actual execution + driver support is execute_sql's job, unchanged here. Added a test.

packages/agami-core/src/tools.py
ashwin-agami: This ranking only decides which metrics get **full** detail inline. \`get_datasource_schema\` ALWAYS returns \`metric_index\` (every metric's name + one-liner), so a metric whose name shares no words with the query is **never hidden** — the LLM sees it exists and can pull it by name via \`metric_names\`. The stopwords just carry no metric-identity signal, so they're dropped from the weak token-overlap path only (exact/substring matches still use them). Core use: stop dumping all 100+ metrics into context on every call. Expanded the comment in 0706488.

PR #38, opened 2026-06-25
ACE-002: backend-portable DB serving (model from DB, logs to DB)

ashwin-agami:
Spec: ACE-002

## Summary

Serve the semantic model + prompt examples from a database and write query logs/feedback to it, so the shared HTTP server (ACE-001) is multi-instance-safe and survives restarts. **Postgres is the prod backend, but the schema + queries are backend-portable** over a thin DB-API adapter — which lets the whole read/write path be tested against SQLite in CI (no DB service), while a future backend stays an additive adapter. YAML remains the source of truth; the DB is built from it; the local file path is unchanged when `AGAMI_DB_URL` is unset.

## Changes (4 slices)

- **`store.py`** — `Store.connect(url)` opens `sqlite://` (stdlib; tests + small self-host) or `postgresql://` (psycopg2, `[server]` extra; prod). One set of SQL with `?` placeholders, adapted to `%s` for Postgres; dict rows on every backend. `run_migrations()` applies ordered `migrations/core/NNN_*.sql` idempotently via a `schema_migrations` tracking table.
- **`migrations/core/001_serving.sql` + `002_runtime.sql`** — per-object tables mirroring the `Organization` tree (org → datasource → subject-area; tables/columns/metrics/entities/relationships/examples/memory/model_version) + `query_executions`/`feedback`. Each object carries its key/structural columns + sizing metadata (`subject_area.table_count`, `model_table.est_row_count`) + a `doc` JSON column for lossless rebuild. Portable DDL (TEXT/INTEGER, app-minted ids — no JSONB/SERIAL). Includes a **table→tool backing map**.
- **`model_store.py`** — the model-loader seam = *produce an `Organization`*: `write_organization` seeds rows from a YAML-loaded org (deploy path); `load_organization` rebuilds the **identical** object so every tool is untouched. `tools._load_org` reads the DB when `AGAMI_DB_URL` is set, else the YAML files.
- **`DbActivitySink`** (one backend-agnostic class, conforms to the `ActivitySink` Protocol by shape) + `get_prompt_examples` DB serving (scope to datasource, word-overlap top-K within a char budget — never the whole library; embeddings tier off by default). `tools` routes execute_sql/log_feedback through the configured sink; the local jsonl path is unchanged.

## Decisions (per the rubric)

- Backend-portable store (Postgres prod / SQLite CI + small self-host); alternate backends are a future adapter, not a rewrite.
- The loader seam is "produce an `Organization`"; the DB loader rebuilds the same object — tools unchanged. (OCR didn't ship a model-loader port; this defines it.)
- One `DbActivitySink` over the portable connection — **not** a Postgres/SQLite pair (it conforms to the port structurally; no inheritance).
- Per-object rows + a `doc` column = lossless rebuild without per-field column sprawl.
- `AGAMI_DB_URL` unset ⇒ the local file path (no regression).

## Verification (SQLite-backed, runs in the gate)

- `uv run dev.py check` green: ruff · **708 tests** · gitleaks. CI installs `[model,server]`.
- Golden parity: a YAML-loaded `Organization`, seeded to the DB and re-loaded from a **fresh connection**, is identical; `get_datasource_schema` head (incl. full/summary/index sizing, `metric_index`, `large_tables`) matches DB vs files.
- `query_executions`/`feedback` rows are written at query time through the single path; a fresh connection reads them (multi-instance-safe).
- A large example library returns a bounded, ranked, scoped set; embeddings off.
- Migrations create all tables on an empty DB, idempotently. `execute_sql.py` (the executor) is unchanged. The no-network contract now covers the new store modules.

## Checklist

- [x] Spec-linked; in scope (no HTTP transport / cloud-neutral config / auth tables — separate specs)
- [x] Ports-and-adapters: DB adapters behind the loader seam + ActivitySink port; tools select by config
- [x] New behavior has tests; no tests weakened; local path unchanged when `AGAMI_DB_URL` unset
- [x] Full local gate green
- [x] No real customer data/names/credentials; the customer's model+logs stay in their own DB (no egress)
- [x] No internal spec/roadmap IDs in repo files (only this `Spec:` line)

## Out of scope

The HTTP transport (ACE-001, merged); cloud-neutral config hardening; the deploy package that runs migrations + loads the model; `save_correction`/governance objects; pgvector / an embedding index. Auth/users tables are app-auth.

> Note: the Postgres backend is exercised via the portable layer (identical SQL + the `%s` adaptation); a live-Postgres integration test would need a DB service, which the portable design deliberately avoids in CI.

### DIFF
--- a/migrations/core/001_serving.sql
+++ b/migrations/core/001_serving.sql
@@ -66,6 +66,9 @@ CREATE TABLE entity (
     PRIMARY KEY (datasource, area, name)
 );
 
+-- WITHIN-subject-area relationships only. Cross-subject-area / cross-datasource relationships are
+-- org-level (Organization.cross_subject_area_relationships) and ride inside organization.doc — they
+-- are rebuilt from there, not stored as rows here (which would need a null area, see the NOTE above).
 CREATE TABLE relationship (
     datasource TEXT NOT NULL,
     area       TEXT NOT NULL,
@@ -84,9 +87,12 @@ CREATE TABLE prompt_example (
     PRIMARY KEY (datasource, id)
 );
 
+-- Domain-context docs. kind='organization' (ORGANIZATION.md) is per-datasource; kind='user'
+-- (USER_MEMORY.md) is install-global and stored once under datasource='' (the empty sentinel),
+-- mirroring the file layout (<artifacts_dir>/<profile>/ORGANIZATION.md vs <artifacts_dir>/USER_MEMORY.md).
 CREATE TABLE memory (
-    datasource TEXT NOT NULL,
-    kind       TEXT NOT NULL,      -- 'organization' (ORGANIZATION.md) | 'user' (USER_MEMORY.md)
+    datasource TEXT NOT NULL,      -- '' for the global user-memory row
+    kind       TEXT NOT NULL,      -- 'organization' (per-datasource) | 'user' (global)
     content    TEXT,
     PRIMARY KEY (datasource, kind)
 );
--- a/migrations/core/001_serving.sql
+++ b/migrations/core/001_serving.sql
@@ -1,0 +1,95 @@
+-- agami-core serving schema — the semantic model, served from the DB instead of local YAML files.
+--
+-- Shape mirrors the in-memory Organization tree (semantic_model/models.py) under an
+-- org -> datasource -> subject-area hierarchy. Each object row carries the structural/key columns
+-- a tool queries on PLUS a `doc` (the object's JSON) so the loader rebuilds the EXACT pydantic
+-- object losslessly — no per-field column sprawl, and YAML stays the source of truth.
+--
+-- Portable DDL: TEXT/INTEGER only, app-minted keys (no SERIAL/JSONB) so the same file runs on
+-- SQLite (tests / small self-host) and Postgres (prod).
+--
+-- table -> tool backing map:
+--   organization, subject_area, model_table, metric, entity, relationship, memory
+--                              -> get_datasource_schema (subject_area.table_count + model_table.est_row_count
+--                                 drive the full/summary/index sizing + large_tables; relationship rows feed
+--                                 execute_sql's internal pre_flight_check; entity.value_pattern feeds the
+--                                 folded identify_entity)
+--   prompt_example             -> get_prompt_examples (scoped by datasource+area, ranked app-side)
+--   model_version              -> the execute_sql trust receipt's version pin
+--   (list_datasources reads the set of `datasource` rows; execute_sql/log_feedback write runtime rows — 002)
+
+CREATE TABLE organization (
+    datasource  TEXT PRIMARY KEY,   -- the profile/datasource id (single-tenant: one org per datasource)
+    org_name    TEXT NOT NULL,
+    description TEXT,
+    doc         TEXT NOT NULL       -- org-level JSON: version, fiscal_year_start_month, glossary,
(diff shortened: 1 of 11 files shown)


migrations/core/001_serving.sql
```diff
@@ -0,0 +1,95 @@
...
+    datasource TEXT NOT NULL,
+    area       TEXT,               -- NULL = org-level cross-datasource example bucket
+    id         TEXT NOT NULL,
+    question   TEXT NOT NULL,
+    doc        TEXT NOT NULL,
+    embedding  TEXT,               -- optional, off by default: deploy-time vector as JSON text (no pgvector)
+    PRIMARY KEY (datasource, id)
+);
+
+CREATE TABLE memory (
```
ashwin-agami, line 93: How does this track both Organization and User memory?

packages/agami-core/src/model_store.py
```diff
@@ -0,0 +1,132 @@
...
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from semantic_model.models import Organization
+from store import Store
+
+# subject-area model tables this module owns (cleared + re-seeded together).
+_MODEL_TABLES = ("relationship", "entity", "metric", "model_table", "subject_area", "organization")
```
ashwin-agami, line 28: Do we need this? And should it be hard coded here?

migrations/core/001_serving.sql
```diff
@@ -64,7 +68,7 @@ CREATE TABLE entity (
 
 CREATE TABLE relationship (
     datasource TEXT NOT NULL,
-    area       TEXT,               -- NULL = cross-subject-area / cross-datasource relationship
+    area       TEXT NOT NULL,
```
ashwin-agami, line 74: where are cross datasource relationships maintained?

migrations/core/001_serving.sql
ashwin-agami: Via the `kind` column — one row per kind: `(datasource,'organization')` and a `user` row. But your question caught a real bug: USER_MEMORY.md is **install-global** (`<artifacts_dir>/USER_MEMORY.md`, no profile) while ORGANIZATION.md is per-datasource — and I was keying both by datasource, which would wrongly make user memory per-datasource. Fixed in 1bdf081: the `user` row is stored once under a global sentinel (`datasource=''`); `organization` stays per-datasource. Schema comment + write/load_memory updated.

migrations/core/001_serving.sql
ashwin-agami: Cross-subject-area / cross-datasource relationships are **org-level** — they live in `Organization.cross_subject_area_relationships`, which rides inside `organization.doc` and is rebuilt from there. The `relationship` table holds **within-subject-area** relationships only (its `area` is NOT NULL — Postgres forbids NULL in a PK column, so org-level rows cant live here). Added a comment on the table saying exactly this (1bdf081).

packages/agami-core/src/model_store.py
ashwin-agami: Yes, we need it — `write_organization` clears a datasource's existing rows before re-seeding so a redeploy reproduces the model instead of appending duplicates / hitting PK conflicts. The hardcoding is a real maintenance coupling (it must track the serving schema), so I added a comment saying so; examples/memory/model_version are re-seeded by their own writers, which is why they're not in the list. If you'd prefer, a later cleanup could derive the table set from a single schema registry rather than two hardcoded places.

PR #42, opened 2026-06-26
OAuth provider — authorize + token + PKCE + JWT issuance (1 of 2)

ashwin-agami:
## Summary

Implements the server's OAuth 2.1 endpoints that the discovery metadata already advertises, so claude.ai's connector can complete its handshake (confirmed: the connector is OAuth-only — no static-key mode). Backed by the existing user store. **PR 1 of 2** into `feature/app-auth`: this is the **provider (issuance)**. PR2 adds the JWT-validating `AuthProvider` + transport wiring + the end-to-end claude.ai smoke.

## Changes

- **`migrations/core/004_oauth.sql`** — `oauth_client` (registered clients) + `oauth_state` (authorization codes bound to their PKCE challenge + redirect + username; single-use + short-lived). Portable DDL.
- **`oauth_server.py`** — `/oauth/register` (minimal RFC 7591 → mints a `client_id`), `/oauth/authorize` (login form → authorization code via the credential check), `/oauth/token` (code + PKCE verifier → self-signed HS256 JWT). Enforces: PKCE S256, single-use + 10-min codes, redirect allow-listing, ≥32-byte signing secret, no secret/token in error bodies.
- **`mcp_http.py`** — routes the three `/oauth/*` endpoints; they're public at the transport layer (pre-auth by definition, boundary-matched), `/mcp` still requires a bearer.
- **`PyJWT`** added to the `[server]` extra.

Ran `/code-review` + a dedicated security pass. The security pass caught **two account-takeover-class must-fix bugs**, both **fixed + regression-tested** in this PR:
- **Open redirect** — redirect validation used `startswith(PUBLIC_BASE_URL)`, so `https://host.com.evil.com` prefix-matched `https://host.com`. Now compares parsed **scheme+host** (same-origin); registered/claude callbacks are exact-match.
- **Reflected XSS** — the login form interpolated OAuth params into HTML attributes unescaped on the password page. Now `html.escape(quote=True)` on every value.

Cleared on review: auth-code single-use ordering, ISO expiry compare, PKCE required (a challenge-less code is unredeemable), HS256-only issuance, parameterized SQL, `/oauth/*` scoping (doesn't open `/mcp`), no password/secret in logs or errors.

## Test plan

`tests/test_oauth_server.py` (14 tests): full PKCE round-trip → verifiable JWT; wrong verifier / reused code / expired code / redirect mismatch → 400; bad creds re-render; **prefix-confusion redirect rejected**; same-origin redirect allowed; **login form escapes params (no XSS)**; missing signing secret fails without burning the code; `/oauth/*` public while `/mcp` 401s. Full gate green (`uv run dev.py check` — 783 tests + ruff + gitleaks).

## Checklist

- [x] New behavior covered by tests; security review done (high lane), 2 must-fix bugs fixed
- [x] Targets `feature/app-auth`

Spec: ACE-005

### DIFF
--- a/migrations/core/004_oauth.sql
+++ b/migrations/core/004_oauth.sql
@@ -14,8 +14,8 @@ CREATE TABLE oauth_client (
 CREATE TABLE oauth_state (
     code           TEXT PRIMARY KEY,        -- the authorization code (secrets.token_urlsafe)
     client_id      TEXT,
-    redirect_uri   TEXT,
-    code_challenge TEXT,                     -- PKCE S256 challenge; verified at token exchange
+    redirect_uri   TEXT NOT NULL,           -- always validated + present before a code is issued
+    code_challenge TEXT NOT NULL,           -- PKCE S256 challenge (required); verified at token exchange
     username       TEXT NOT NULL,           -- the authenticated principal the code stands in for
     expires_at     TEXT NOT NULL,           -- short-lived; expired codes are rejected
     used           INTEGER NOT NULL DEFAULT 0,  -- single-use: set on first successful exchange
--- a/migrations/core/004_oauth.sql
+++ b/migrations/core/004_oauth.sql
@@ -1,0 +1,23 @@
+-- agami-core OAuth provider schema — the server's own authorize/token flow.
+--
+-- oauth_client: clients that registered (claude.ai via the minimal RFC-7591 endpoint). Minimal by
+-- design — full DCR (client auth, rotation) is deferred. oauth_state: one row per issued
+-- authorization code, bound to its PKCE challenge + redirect_uri + the authenticated username, with
+-- a short expiry and a single-use flag. Keyed by app-minted values (no SERIAL), like the rest.
+
+CREATE TABLE oauth_client (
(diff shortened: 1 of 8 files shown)


ashwin-agami:
Addressed all 6 Copilot suggestions (commit 5ef6aa6):

1. **Auth-code race** — `/oauth/token` now burns the code atomically (`UPDATE … WHERE used=0` + rowcount check), so two concurrent exchanges can't both issue a JWT.
2. **register input** — `redirect_uris` is validated as a list of strings; malformed input returns a 400 OAuth error, not a 500.
3. **PKCE mandatory** — `/oauth/authorize` rejects a missing `code_challenge` up front instead of persisting an unredeemable code.
4. **test** — the bad-credentials test now carries a valid PKCE challenge so it exercises the credential path.
5. **exact `/oauth/*` skip** — the public-path check is exact-match for the OAuth endpoints (boundary-match stays only for the `.well-known/{rest:path}` discovery routes), so a future `/oauth/token/...` route can't inherit public access.
6. **schema** — `oauth_state.redirect_uri` and `code_challenge` are now `NOT NULL` (the request logic guarantees both).

Added regression tests for the PKCE-required, malformed-register, and exact-match cases. Full gate green (785 tests + ruff + gitleaks).

PR #43, opened 2026-06-26
JWT-validating AuthProvider + transport wiring (2 of 2)

ashwin-agami:
## Summary

The **consumption side** of the OAuth flow: validates the JWT the provider (PR1, #42) issues and wires it into the transport, so the hosted server trusts only tokens it issued. **PR 2 of 2** into `feature/app-auth` — this completes the app-auth feature; after it the whole branch gets the end-to-end smoke and merges to main as one reviewed PR.

## Changes

- **`oauth_server.JwtAuthProvider`** — validates the HS256 JWT: algorithm pinned to `HS256` (no `alg=none`/confusion), requires `sub`/`exp`/`iss`, verifies the issuer == `PUBLIC_BASE_URL`, fails closed (any bad/expired/forged token → `None` → 401). Conforms structurally to the `AuthProvider` port.
- **`mcp_http._build_auth_provider()`** — selects the JWT validator when `AGAMI_SIGNING_SECRET` is set (only this server's tokens are accepted), else the bearer-presence default (the no-secret local fallback); threaded through the auth middleware. The choice is frozen at `build_app` time — a deployment with the secret set never falls back to presence.
- Refreshed the now-stale "later feature" docstrings (authorize/token/JWT exist now).

Ran `/code-review` + a dedicated security pass — both came back clean. The security pass verified the validator is airtight against the classic JWT attacks (alg-confusion / `alg=none`, signature, expiry, issuer) and fails closed. Per its suggestion, added explicit tests asserting `alg=none`, `alg:HS512`, and wrong-issuer tokens are all rejected.

## Test plan

`tests/test_oauth_server.py`: `JwtAuthProvider` accepts an issued token, rejects junk / wrong-secret / `alg=none` / `alg:HS512` / wrong-issuer; **end-to-end** register→authorize→token→use the JWT as Bearer on `/mcp` → `tools/list` returns exactly the 5 product tools; a non-JWT bearer is rejected when the secret is set; presence still works with no secret. Full gate green (790 tests + ruff + gitleaks).

## Checklist

- [x] New behavior covered by tests; security review done (high lane), both reviews clean
- [x] Targets `feature/app-auth`

Spec: ACE-005

### DIFF
--- a/packages/agami-core/src/mcp_http.py
+++ b/packages/agami-core/src/mcp_http.py
@@ -4,9 +4,9 @@
 Advertises the **same** shared `tools.TOOLS` registry as the stdio entrypoint, but over a network
 endpoint a remote client (claude.ai) can reach — plus the OAuth-discovery surface and a bearer
 auth shim. This is the network product: unlike the stdio harness it binds a port, so it carries
-auth — an unauthenticated request gets a `401` + `WWW-Authenticate` challenge, which is what
-triggers the client's OAuth flow. The authorize page + real identity are a later concern; here we
-emit the challenge and require a bearer token's presence (the OSS `AuthProvider` default).
+auth — an unauthenticated request gets a `401` + `WWW-Authenticate` challenge, which triggers the
+client's OAuth flow against the authorize/token endpoints (see `oauth_server`). The issued JWT then
+gates `/mcp`; with no signing secret configured the OSS bearer-presence default applies instead.
 
 Requires the **[server]** extra (the MCP SDK + ASGI stack). `PUBLIC_BASE_URL` must be set
 explicitly — it backs the discovery documents + the `WWW-Authenticate` resource URL and cannot be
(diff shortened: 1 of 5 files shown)


ashwin-agami:
Addressed all 3 Copilot suggestions (commit 4083cf9):

1. **Typed `auth`** — `_AuthMiddleware.__init__` now takes `auth: ports.AuthProvider`, making the contract explicit.
2. **`sub` validation** — `JwtAuthProvider` rejects a non-string or blank `sub` (returns `None`) instead of constructing a `Principal` from a malformed identity.
3. **No silent downgrade** — `_build_auth_provider` now treats `AGAMI_SIGNING_SECRET` being *present* (even empty/weak) as JWT-mode intent and validates it at `build_app` (fail fast), rather than falling back to presence auth on a misconfigured hosted deploy. Only an **absent** var selects the local presence default.

Added tests: a present-but-empty / too-short secret fails the build; a non-string / blank `sub` is rejected. Full gate green (792 tests + ruff + gitleaks).

ashwin-agami:
Addressed the follow-up (commit pushed): the `base_url` fixture now clears `AGAMI_SIGNING_SECRET` so the bearer-presence tests are deterministic regardless of a developer's ambient env. Verified the presence suite passes even with the secret exported.

PR #46, opened 2026-06-26
OIDC: Microsoft (tenant-pinned) + provider-binding + public-demo (2 of 2)

ashwin-agami:
## Summary

Completes OIDC social login. **PR 2 of 2** into `feature/oidc`: adds **Microsoft** (tenant-pinned), the **provider-binding** that closes the IdP-confusion risk PR1 flagged, and the opt-in **public-demo signup** mode. After this, `feature/oidc` → main.

## Changes

- **Microsoft provider** — discovery from a **pinned** tenant (`AGAMI_OIDC_MICROSOFT_TENANT`; `common`/`organizations`/`consumers` are refused — the tenant is the trust boundary). `verify_id_token` returns `Identity(email, subject)` and applies `email_verified` **per provider**: required for Google (consumer IdP), not for tenant-pinned Microsoft (which often omits it). `preferred_username` is never used as identity.
- **Provider-binding (the IdP-confusion fix)** — `006_oidc_identity.sql` adds `oidc_provider` + `oidc_subject`. A user is bound to **one** provider (set at onboarding) and **one** subject (TOFU on first login, no-clobber). A token from another IdP, or a different account at the same IdP, is **refused** rather than resolved to the victim.
- **Public-demo signup** — `AGAMI_PUBLIC_SIGNUP` (default **off**, fail-closed) lets an unknown verified email self-provision a `status='demo'` user — for a dedicated "Try for free" instance whose data is non-sensitive. Demo users get no elevated capability.
- **README** — per-provider env, the tenant pin, the demo flag, and the hosted-only **egress** note.

Ran `/code-review` + a dedicated security pass. The **security pass came back clean** — verified the provider-binding (provider + subject) closes IdP confusion, the MS tenant pin is unbypassable (and is the trust for relaxing `email_verified`), demo-signup is fail-closed + un-elevated + can't hijack an existing user, all failures collapse to a generic 403, no secret/PII leak, no SSRF. Acted on the code review's findings: a misconfigured Microsoft tenant now returns a clean **400 (not a 500)** at the HTTP layer, the demo `create_user` is guarded against a UNIQUE race, and `_resolve_oidc_user` is typed.

## Test plan

`tests/test_oidc.py` (mocked IdP, real RS256, no egress): Microsoft flow **without** `email_verified`; Google **rejected** without it; **unpinned tenant refused**; **provider mismatch → reject** (IdP confusion); **subject TOFU** bind-then-enforce; demo signup **on** creates a demo user, **off** rejects; disabled user rejected; misconfigured-tenant `/oauth/oidc/start` → 400. Full gate green (818 tests + ruff + gitleaks).

## Checklist

- [x] New behavior covered by tests; security review done (high lane), clean
- [x] Targets `feature/oidc`

Spec: ACE-006

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,11 +187,28 @@ is entirely environment variables:
 | `AGAMI_DB_URL` | yes (server) | The store: `postgresql://…` in prod, `sqlite://…` for a small/local run. `APP_DATABASE_URL` is accepted as an alias for the cloud-platform convention (`AGAMI_DB_URL` wins if both are set). Unset ⇒ the local file path. |
 | `PUBLIC_BASE_URL` | yes (server) | Backs OAuth/MCP discovery + the `WWW-Authenticate` resource URL. Set it explicitly — it can't be auto-detected behind a proxy/LB; the server fails fast at startup if it's missing. |
 | `AGAMI_ORG_ID` | no | The single configured org id (default `local`). The server is single-tenant by default. |
+| `AGAMI_SIGNING_SECRET` | yes (auth) | ≥32-byte secret the server signs its own session JWTs with. When set, the server validates real tokens (the password / OIDC login flow); unset ⇒ the bearer-presence local default. |
 
 All serving state lives in Postgres — a fresh instance with only `AGAMI_DB_URL` serves identically,
 so the server survives restarts and stateless platforms. The serving path is **LLM-free and
 zero-egress by default**: the client is the brain, `execute_sql` runs SQL against your own database,
 and the other tools just read the model. Nothing leaves your environment.
+
+### Optional: social login (OIDC)
+
+"Sign in with Google / Microsoft" is **off by default**. Configure a provider's client id/secret to
+enable it (the option is hidden when unset; username/password still works):
+
+| Variable | Provider | Notes |
+|----------|----------|-------|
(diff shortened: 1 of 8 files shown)


ashwin-agami:
Addressed the Copilot finding (concurrent first-login subject-bind race): the resolver now does **bind-then-verify** — it attempts the no-clobber bind, then re-reads and requires the stored subject is ours. If a concurrent login bound a different subject first, our UPDATE no-ops, the re-read mismatches, and we **reject** rather than completing login against another binding. Gate green (818 tests).

ashwin-agami:
Addressed all 4 Copilot re-review comments (latest commit):
1. **Backfill** — migration 006 now sets `oidc_provider='google'` for pre-existing passwordless users (the only pre-binding provider) so the new gate can't lock them out on upgrade.
2. **Integrity-only catch** — the demo `create_user` now swallows only UNIQUE/integrity collisions (portable MRO check) and **re-raises** real failures, so a datastore outage surfaces as a 500, not a masked 403.
3. **`sub` validation** — `verify_id_token` rejects a missing/blank/non-string `sub` (it's the long-term binding key). Test added.
4. **Docstring** — `available_providers` now correctly states a misconfigured MS tenant yields a clean 400.
Gate green (819 tests).

PR #50, opened 2026-06-26
Teammate self-onboarding (deployment-wide auth method)

ashwin-agami:
## Summary

Completes teammate onboarding. An admin creates a teammate by **email + name** (a *pending* user who
can't yet sign in); this lets them finish setting up — following a **deployment-wide auth method**,
uniform for everyone, set by what's configured:

- **OIDC deployment** (a Google/Microsoft client is configured): a pending teammate signs in with that
  provider at the connector — the IdP verifies their email and **binds the account on first login**. No
  link to share.
- **Password deployment** (no OIDC): the admin **copies a setup link** from the Users tab and shares it
  out-of-band; the teammate opens it and **sets their own password**. The link is a signed, time-boxed
  token (no new table) and is **single-use** — it stops working once the account is set up.

Because a deployment offers one method, a user can never hold two credentials — so there's no per-user
choice and no lock-juggling. Login surfaces show **only the configured method** (the admin keeps a
password **break-glass** fallback on `/admin/login`).

## Changes

- **`user_store.py`** — `claim_pending_oidc` / `claim_pending_password`: guarded
  `UPDATE … WHERE password_hash IS NULL AND oidc_provider IS NULL` (race-safe, mirrors the ACE-006 bind).
- **`oauth_server.py`** — `_resolve_oidc_user` gains a *pending* branch (bind on first OIDC login, then
  re-read and require the binding is ours); `login_body_html` renders only the configured method, and
  `/oauth/authorize` **refuses a password POST server-side** in an OIDC deployment.
- **`onboarding.py`** (NEW) — the setup token (purpose-marked HS256, no new table) + the `/claim` setup
  page/handler. Unforgeable, single-use, never overwrites a claimed account; hashes only after the token
  + pending checks.
- **`admin.py`** — the Users tab shows a copy-able **setup link** for pending users, **only** in a
  password deployment (the page is session-gated).
- **`mcp_http.py`** — `/claim` routes + the bearer public-skip (boundary-matched). Docs + previews.

## Test plan

- `tests/test_onboarding.py` — token round-trip + forged/expired/wrong-purpose rejection; `/claim` sets a
  password for a pending user; **single-use** (replay refused, original password still works, attacker's
  rejected); short/bad-token refused; can't overwrite a claimed/admin account; roster shows the setup
  link for pending users **only** in a password deployment.
- `tests/test_oidc.py` — pending user binds on first OIDC login; a password user is never a pending OIDC
  claim; **first-claim-locks** both directions; connector login shows only the configured method; a
  password POST is refused server-side in an OIDC deployment.
- Gate green: ruff + ruff-format + full pytest (890) + gitleaks; patch coverage 100% (onboarding 96% — the
  3 uncovered lines are defensive store-unavailable guards).
- Reviewed with `/code-review` + `/security-review` (new public credential surface); findings addressed.

## Checklist

- [x] Tests added and passing; gate green
- [x] No real customer names / data / credentials — neutral placeholders only
- [x] No new network egress (OIDC reuses `oidc.py`, the existing single egress module)
- [x] Flat access preserved (no role/permission column)
- [ ] Manual end-to-end smoke behind a Cloudflare tunnel — OIDC bind + the setup-link password path

Spec: ACE-011

### DIFF
--- a/packages/agami-core/README.md
+++ b/packages/agami-core/README.md
@@ -94,6 +94,10 @@ the deployment's **single auth method** — uniform for everyone, set by what yo
 The login surfaces show only the configured method (the admin keeps a password **break-glass** fallback
 on `/admin/login`).
 
+> **Trust note.** OIDC onboarding binds the account to whoever first proves the teammate's email at the
+> configured IdP — so add a teammate by an email the *right* person controls there. The setup link and
+> the other pre-auth endpoints aren't rate-limited in-process; put them behind your proxy/LB if exposed.
+
 ### Local end-to-end test (HTTPS via a tunnel)
 
 OAuth and the `Secure` admin cookie need HTTPS, so expose the local server through a tunnel:
--- a/packages/agami-core/README.md
+++ b/packages/agami-core/README.md
@@ -79,9 +79,20 @@ A deployment is one host (`PUBLIC_BASE_URL`). Everything lives under it:
   refused — the pin closes IdP-confusion). An `/mcp` bearer token is useless here (different
   credential). Unset `AGAMI_ADMIN_USERNAME` ⇒ the admin console is disabled entirely.
 
-The admin adds a teammate by **email + name**; the teammate then signs in at the connector. (Letting
-that teammate choose their own sign-in method on first login — Google/Microsoft or a self-set
-password — is the next increment.)
+### Onboarding a teammate
+
+The admin adds a teammate by **email + name** (a *pending* user). How they finish setting up follows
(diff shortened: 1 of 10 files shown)


ashwin-agami:
Thanks @copilot — fixed in 3e2e07c. You're right that the JWT `sub` is base64url-readable; the docstring now says the token is **signed (unforgeable)** rather than "opaque", and clarifies the no-enumeration property comes from the handler returning the same generic page regardless of token/user state — not from the token hiding who it's for. Doc-only; no behavior change.

PR #51, opened 2026-06-27
Admin activity views — Tool calls + Sessions + design polish

ashwin-agami:
## Summary

Adds the admin **activity** surface to the hosted MCP server — two read-only lenses on one `tool_calls`
log — plus a design polish to the shared admin/auth shell.

- **Tool calls** (`/admin?tab=calls`) — **audit-grade**: every MCP tool call, newest first, with **who**
  (the authenticated user), the tool, datasource, and for a query the SQL / row count / latency / status.
  Always accurate (the server observes it directly).
- **Sessions** (`/admin?tab=sessions`) — **best-effort**: those queries grouped into a conversation, each
  opening to its queries with the natural-language **question**. The MCP protocol carries neither the
  user's question nor a conversation id, so Claude self-reports them (a `user_question` param + a
  `thread_id` it reuses); when it does, you get grouping + the question, and when it doesn't the view
  degrades to ungrouped. These two fields are clearly labelled self-reported.
- **Design polish** in `ui.py` (so login / admin / onboarding all benefit): self-hosted **Inter** (no
  font-CDN call), refined buttons/tables, full-width + mobile, and **browser-local** human times.

The read-only **model view** was split to its own spec (it's a bigger surface).

## Changes

- **`migrations/008_tool_calls.sql` + `model_store.py` + `contracts.py`** — the `tool_calls` table, a
  `record_tool_call` writer, and the `list_tool_calls` / `list_sessions` read helpers (sessions grouped
  in Python on `thread_id`, with NULL → ungrouped singleton).
- **`mcp_http.py`** — a "record every tool call" hook in `_call_tool` (timed, best-effort, never breaks
  or alters a tool), and the **actor plumbing**: the validated `Principal` rides the ASGI scope state and
  a contextvar set in the raw-ASGI `/mcp` endpoint (same task as the dispatch) carries it into the
  handler — the actor is the bearer subject, never a tool argument.
- **`tools.py`** — `record_tool_call` (derives success/row_count from the result; never raises), the
  `execute_sql` self-report params, and a server-instructions line asking Claude to send them.
- **`admin.py`** — the Tool calls + Sessions tabs (flat table / grouped, each with a detail drawer),
  wired into the console; everything escaped; admin-gated; no secret in the output.
- **`ui.py` + `static/fonts/`** — the shared polish + self-hosted Inter woff2 (latin subset).

## Test plan

- `tests/test_tool_calls.py` — the recorder (success/error/exception derivation, self-report fields,
  non-JSON tolerance, best-effort never raises) + the **end-to-end actor gate** (an authenticated `/mcp`
  call logs the JWT actor) + a raising tool still logged as an error.
- `tests/test_admin_activity.py` — `list_sessions` grouping / degradation / all-null-latency; the two
  tabs show the actor + escape SQL/question; admin-gated; no `password_hash` leak.
- `tests/test_ui_polish.py` — self-hosted Inter (no font CDN), the time script, full-width, font served.
- Gate green: ruff + ruff-format + full pytest (909) + gitleaks; patch coverage strong (the 2 uncovered
  lines are defensive logging-failure guards).
- Reviewed with `/code-review` + `/security-review` (auth/data path) — 0 must-fix; findings addressed
  (dropped a throwaway mock; hardened drawer DOM ids; added an edge test; documented log retention).

## Checklist

- [x] Tests added and passing; gate green
- [x] No real customer names / data / credentials — neutral placeholders only
- [x] No new network egress (logging is local; Inter is self-hosted — *reduces* egress)
- [x] Flat access preserved; views are read-only + admin-gated
- [ ] Manual smoke behind a Cloudflare tunnel — confirm the log fills, and **validate whether Claude
  echoes `thread_id` + fills `user_question`** (the go/no-go on session grouping)

Spec: ACE-008


ashwin-agami:
Thanks @copilot — both fixed in 5fe231e:

1. **Cross-actor session blending** (`model_store.list_sessions`): the grouping key is now `(actor, thread_id)` — since `thread_id` is self-reported/untrusted, two users colliding on one can no longer merge into a single, misattributed session. Added a regression test asserting the same `thread_id` across different actors yields two correctly-attributed sessions.
2. **`render_previews.py`**: switched to `tempfile.mkstemp` (atomic) from the race-prone `mktemp`.

Gate green, 910 tests.

PR #54, opened 2026-06-28
Turn-level observability — correlation_id binds a question's agent refinements

ashwin-agami:
**Spec: ACE-015**

## Summary
Adds the **turn** level to the activity log — the deferred middle of ACE-008's `thread_id` (conversation) ▸ **`correlation_id` (turn)** ▸ step. One user question fans out into N agent sub-queries; this binds them, so the admin Sessions drawer reads *"User asked X → 2 queries"*. Grounded in real data (a meridian session showed the gap **and** that Claude drifts `user_question` onto later refinements).

## Changes
- **`correlation_id`** — migration `009` adds the nullable column; `ToolCallRecord` + the DB sink + `_TOOL_CALL_COLS` carry it; `record_tool_call` captures `args.get("correlation_id")`.
- **Self-reported, per turn** — `execute_sql`'s `inputSchema` gains a `correlation_id` param, and `SERVER_INSTRUCTIONS` now tells Claude to keep `user_question` **verbatim** across a turn (refinement goes only in `raw_query`) and mint one `correlation_id` per user question. (The server can't mint it — the MCP protocol carries no turn boundary.)
- **Turn grouping** — `model_store._group_turns` folds a session's calls into turns by `correlation_id`; the turn's question is the **earliest** call's `user_question` (drift-proof); a call with no `correlation_id` is its own singleton turn. `_session_drawer` renders one *"User asked … · N queries"* block per turn with each refinement's `agent_query` + SQL.
- Preview seeds a 2-query turn; README documents the turn level.

## Safety / degradation
Best-effort + graceful: no `correlation_id` → reads like the old flat list, never errors. Read-only + admin-gated; every self-reported field escaped; `correlation_id` is never rendered into HTML (used only as a grouping key); no new egress.

## Test plan
`tests/test_admin_activity.py`: `correlation_id` round-trips; the `inputSchema` exposes it; `SERVER_INSTRUCTIONS` carries the verbatim + per-turn guidance; turns group by `correlation_id`; **the drift case** (2nd call's question differs → earliest wins); **degradation** (no `correlation_id` → singleton turns); the drawer renders *"User asked"* + both `agent_query` refinements; the existing XSS-escape test still covers the sessions tab.

## Checklist
- [x] `uv run dev.py check` green — **953 tests**, ruff, gitleaks
- [x] **100%** patch coverage (`uv run dev.py cover`)
- [x] `/code-review` (2 finder angles) — **0 findings**; Agami rubric 0 must-fix
- [x] Read-only + admin-gated; self-report fields escaped; migration portable; no new egress
- [x] Decisions recorded in the spec; **manual smoke pending** (validate Claude supplies `correlation_id` via the live Claude connection)

### DIFF
--- a/migrations/core/009_tool_calls_correlation.sql
+++ b/migrations/core/009_tool_calls_correlation.sql
@@ -1,0 +1,7 @@
+-- Add the **turn** level to the tool-call log (ACE-015). ACE-008 captured `thread_id` (the whole
+-- conversation) and the per-call `user_question`/`agent_query`; this adds `correlation_id` — the turn,
+-- i.e. the ONE user question whose answer fanned out into several agent sub-queries. Self-reported by
+-- Claude (the MCP protocol carries no turn boundary, so the server can't mint it), nullable and
+-- best-effort like the other self-report columns; the Sessions view groups a session's calls by it and
+-- degrades to ungrouped when absent. Portable (runs on SQLite + Postgres unchanged).
+ALTER TABLE tool_calls ADD COLUMN correlation_id TEXT;
--- a/packages/agami-core/README.md
+++ b/packages/agami-core/README.md
@@ -105,12 +105,14 @@ The admin console also has two read-only activity tabs:
 - **Tool calls** — *every* MCP tool call, newest first: who (the authenticated user), the tool,
   datasource, and for a query the SQL, row count, latency, and status. This is **audit-grade** — the
   server observes it directly, so it's always accurate.
-- **Sessions** — those queries grouped into a conversation, each opening to its queries with the
-  natural-language **question**. This is **best-effort**: the MCP protocol carries neither the user's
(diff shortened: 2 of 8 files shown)


ashwin-agami:
Thanks @copilot:
1. **Self-reported label** — fixed in `42d99d7`: the turn question now renders `User asked <q> · self-reported`, consistent with the rest of the activity log.
2. **Read path 500s if migration 009 isn't applied** — fair, but this is the *established* convention: every migration (incl. ACE-008's `008` that created `tool_calls`) requires the deploy to apply it before the new code reads the new column; making one column's read path defensive would diverge from the other 13. The robust fix you suggest — auto-applying migrations on startup/open-store — is a good general improvement that belongs in the **deploy** work (ACE-009, the `agami deploy` package), not this slice. Logged it there.

PR #65, opened 2026-07-01
OCR-032: publish agami-core to PyPI on release (trusted publishing)

ashwin-agami:
## Summary

Publishes `agami-core` to PyPI on release via **trusted publishing** (OIDC), so a marketplace `/agami-connect` can `pip install "agami-core[model]"` from the index instead of the git fallback. OCR-031's `sm` installer already tries the index before git, so the moment a version lands on PyPI every marketplace install upgrades **with no code change**. This is the packaging companion to the shipped ACE-025 (which publishes the server *image* to GHCR) — same trigger shape, same `workflow_dispatch` smoke-test, same "no long-lived secret" posture, for the Python *package*.

## Changes

- **`.github/workflows/release-pypi.yml`** (new) — builds the sdist + wheel from `packages/agami-core` (`python -m build`) and publishes via `pypa/gh-action-pypi-publish`. `permissions: id-token: write` mints the OIDC token per-run — **no PyPI API token is stored in the repo**. A published `release` publishes to PyPI; `workflow_dispatch` publishes to **TestPyPI** so the pipeline can be smoke-tested first. All actions SHA-pinned.
- **`tests/test_release_pypi_workflow.py`** (new) — locks the invariants CI can't otherwise see: `id-token: write` at job scope, **no `password:`/`secrets.*`** on the publish path, the build sources `packages/agami-core`, and the release→PyPI / dispatch→TestPyPI split.

## One-time manual prerequisite (👤, before the first real publish)

Register the `agami-core` project on PyPI + add a GitHub Actions **trusted publisher** (owner `AgamiAI`, repo `agami-core`, workflow `release-pypi.yml`), and the same on **TestPyPI** to enable the `workflow_dispatch` smoke test. Like ACE-025's "make the GHCR package public" — the workflow is built + TestPyPI-smoke-testable now; the real publish waits on it. There is no secret to manage (OIDC).

## Checklist

- [x] `Spec: OCR-032`
- [x] Full gate green locally — 1024 tests pass, ruff lint + format clean, gitleaks clean
- [x] New behavior ships with a test (`test_release_pypi_workflow.py`, 4 assertions)
- [x] No secrets / no customer data; trusted publishing is tokenless
- [ ] 👤 PyPI + TestPyPI trusted-publisher registration (owner, before first publish)

Spec: OCR-032

### DIFF
--- a/.github/workflows/release-pypi.yml
+++ b/.github/workflows/release-pypi.yml
@@ -1,0 +1,53 @@
+name: Release PyPI
+
+# Builds the agami-core sdist + wheel from packages/agami-core and publishes them to PyPI via
+# TRUSTED PUBLISHING (OIDC) -- so a marketplace `/agami-connect` can `pip install "agami-core[model]"`
+# from the index instead of the git fallback. OCR-031's `sm` installer already tries the index before
+# git, so the moment a version lands on PyPI every marketplace install upgrades with no code change.
+# The version published is packages/agami-core/pyproject.toml `version` (a release tag must match it).
+#
+# Manual prerequisite (ONE-TIME, no secret to manage): register the `agami-core` project on PyPI and add
+# a GitHub Actions "trusted publisher" -- owner AgamiAI, repo agami-core, workflow release-pypi.yml,
+# environment (none). Add the same trusted publisher on TestPyPI to enable the workflow_dispatch smoke
+# test below. Trusted publishing uses the OIDC `id-token` minted per-run (see permissions) -- there is
+# NO PyPI API token stored in the repo.
+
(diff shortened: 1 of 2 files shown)


ashwin-agami:
Addressed Copilot's comment: added a release-only guard (`Verify release tag matches package version`) that compares `GITHUB_REF_NAME` (leading `v` stripped) to the `packages/agami-core/pyproject.toml` version and fails fast on a mismatch, so a mistagged release can't permanently mislabel a PyPI version. Locked with a test assertion (`test_release_guards_tag_matches_version`). `4c5dccc`.

PR #67, opened 2026-07-01
OCR-033: harden sm as the install chokepoint + migrate the two straggler scripts

ashwin-agami:
## Summary

Hardens `sm` as the single install chokepoint and migrates the two runtime scripts that still hand-rolled the pre-OCR-030/031 patterns — fixing four bugs found in the 0.3.2 marketplace run-through (three hard blockers). Real-DB credential promotion and `/agami-serve` crashed on a marketplace install; `sm` was fooled by the bundled `semantic_model` stub and dead against an externally-managed Homebrew Python.

## Changes

- **`plugins/agami/scripts/sm`** (chokepoint hardening):
  - **#5** `_imports_ok` now probes the real entrypoint `import semantic_model.cli` (the OCR-030 `lib/` stub's bare `__init__` no longer satisfies the readiness check), from a neutral cwd; and the final `exec … -m semantic_model.cli` runs after `cd /` (+ `PYTHONSAFEPATH=1`) so a `semantic_model` in the caller's cwd can't shadow the installed package.
  - **#2** `_pip_try` adds a PEP-668 last-resort tier (`--break-system-packages`) so an externally-managed interpreter (Homebrew) can still be equipped.
- **`plugins/agami/scripts/promote_credentials.py`** (**#1**): bare `import agami_paths` (off the script dir) → `_agami_lib.ensure_importable()` resolver. This was crashing credential promotion on a real-DB marketplace onboarding.
- **`plugins/agami/scripts/setup_desktop_mcp.py`** (**#1 + #8**): same resolver; the install now **delegates to `sm install`** (drops the hardcoded `packages/agami-core[model]` path that doesn't exist in a marketplace install), skips when the package is already importable, and `find_interpreter` prefers an interpreter that already has agami-core; `read_version` derives from the cache-dir name.
- **Tests**: new `test_marketplace_script_imports.py` (both scripts run in a `scripts/`+`lib/`-only layout with no `ModuleNotFoundError`; no `packages/` install); extended `test_sm_install_resolution.py` (PEP-668 tier + `.cli`/path-isolation); updated `test_setup_desktop_mcp.py` to the sm-delegation API.

## Test plan

- `uvx --with-editable "packages/agami-core[model,server]" --with pytest pytest tests/` → **1031 passed**; ruff + format + gitleaks clean.
- Each acceptance criterion has a gated test (#1/#2/#5/#8). Real end-to-end install stays a manual check (network/GUI, not CI-gated — same posture as OCR-031).

## Checklist

- [x] `Spec: OCR-033`
- [x] Full gate green locally (1031 tests, ruff, gitleaks)
- [x] New behavior ships with tests; no test weakened (setup_desktop tests adapted to the new delegation API + coverage added)
- [x] No secrets / no customer data
- [x] Decisions logged in the spec (`sm install` = the chokepoint; path-isolation over stub-rename)

Spec: OCR-033

### DIFF
--- a/plugins/agami/scripts/promote_credentials.py
+++ b/plugins/agami/scripts/promote_credentials.py
@@ -31,6 +31,15 @@ import re
 import sys
 from pathlib import Path
 
+# agami_paths lives in the agami-core package; the resolver puts it on the path in every layout
+# (pip-installed / the plugin's bundled lib / a dev checkout) with no pip required. A bare
+# `import agami_paths` off the script's own dir breaks on a marketplace install, where agami_paths
+# lives in lib/, not next to this script (mirrors connect_resolve.py).
+from _agami_lib import ensure_importable  # noqa: E402
+
+ensure_importable()
+import agami_paths  # noqa: E402
+
 # Placeholder tokens the Phase 0a templates ship with — their presence means the user
 # hasn't filled the template in yet. Mirrors the grep guard the skill used inline.
 _PLACEHOLDER_RE = re.compile(
@@ -120,9 +129,6 @@ def promote(agami_dir: Path) -> tuple[str, int]:
 
 
 def main() -> int:
-    import sys as _sys
-    _sys.path.insert(0, str(Path(__file__).resolve().parent))
-    import agami_paths
     agami_paths.bootstrap()
     p = argparse.ArgumentParser(description=__doc__)
     p.add_argument("--agami-dir", default=None,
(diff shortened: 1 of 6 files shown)


ashwin-agami:
Both comments addressed in the latest commit:

1. **`mcp_harness` importable ≠ `[model]` deps present** — right, base `agami-core` has no deps. Removed the divergent skip-check entirely and now **always delegate to `sm install`** (idempotent — no-ops when ready), then verify the **real** `semantic_model.cli` entrypoint (which pulls pydantic/sqlglot/pyyaml). `find_interpreter`'s preference pass uses the same bar. Locked with `test_ensure_package_installed_checks_model_entrypoint`.
2. **`bash` hard-requirement** — added a `shutil.which("bash")` preflight that raises a clear, actionable RuntimeError (with a `pip install` fallback) instead of an opaque FileNotFoundError. Locked with `test_ensure_package_installed_raises_without_bash`.

PR #68, opened 2026-07-01
OCR-034: agami-connect sample determinism + model-explorer + preflight/routing polish

ashwin-agami:
## Summary

The skill/UX half of the 0.3.2 run-through findings (#3/#4/#6/#7): agami-connect leaned on the model's *discretion* where it should be deterministic, and didn't surface the model it just built. No engine code — SKILL.md prose + one small additive `connect_resolve.py` change.

## Changes

- **#3 — preflight stops inventing `main`** (`connect_resolve.py`): adds a `profile_source` field (`explicit|env|active_profile|default`) and emits `profile: null` **only** when `default` AND `next == "bootstrap"` (a real `[main]` section on ready/promote is never hidden; a named-but-unonboarded profile keeps its name). The Phase-0 narration ([SKILL.md:82]) now says *"No credentials yet — first-time setup"* with no invented name.
- **#6 — codify the sample carve-outs** (agami-connect SKILL): 6B ("watch it build") now explicitly **skips the prune / org-description / doc-intake prompts** and builds silently over all sample tables; fixed the `:323`-vs-6B contradiction (the copy path 6A skips introspect/enrich/seed, 6B runs them).
- **#7 — 6B opens the model-explorer** (agami-connect SKILL): after the model validates, 6B renders `/agami-model` in **browse mode** so the user actually sees what was built (the whole point of "watch it build") — the NL→SQL examples-validation page stays skipped for the sample.
- **#4 — model-introspection guardrail** (agami-query SKILL): strengthened the "never hand-read the YAML" rule to also forbid ad-hoc `python -c` model dumps and route *"show me the model"* → `sm model-tree` / `/agami-model` (the run-through's hand-rolled dump crashed + leaked a traceback).
- **Tests**: extended `test_connect_resolve.py` (bootstrap→null+source, ready-with-`[main]` not nulled, named-profile-bootstrap keeps name, `profile_source` precedence); new `test_skill_guardrails.py` (decidable "the skill says X" checks for #4/#6/#7 + the #3 narration).

## Test plan

- `uvx --with-editable "packages/agami-core[model,server]" --with pytest pytest tests/` → **1040 passed**; ruff + format + gitleaks clean.
- Each acceptance criterion has a gated test (#3 behavioral; #4/#6/#7 skill-text). The "watch it build opens the explorer" + "first-time setup narration" E2E is a manual check (S).

## Checklist

- [x] `Spec: OCR-034`
- [x] Full gate green (1040 tests, ruff, gitleaks)
- [x] New behavior ships with tests; no test weakened (connect_resolve tests updated for the null-on-bootstrap contract + coverage added)
- [x] No secrets / no customer data (`agami-example` / `demo` / `staging` placeholders only)
- [x] Decisions logged in the spec (6B opens the explorer; NL→SQL page stays skipped; preflight signals unresolved rather than inventing `main`)

Spec: OCR-034


ashwin-agami:
Addressed: both skill-file reads in `test_skill_guardrails.py` now pass `encoding="utf-8"` so the guardrail assertions are locale-deterministic (the files are repo-authored UTF-8 Markdown).

PR #71, opened 2026-07-02
OCR-035: fix sm discover render_prune crash on real installs + hygiene

ashwin-agami:
## Summary

Fixes a real bug found in a post-sprint hygiene audit: **`sm discover` crashes on any pip/marketplace install** — plus a few cleanups from the same audit. No behavior change beyond the discover fix.

## The bug (🔴)

`sm discover` (the Phase 1.6 table-prune page in a real-DB onboarding) → `python -m semantic_model.cli discover` → `cmd_discover` did `import render_prune` from `Path(__file__).parent.parent`. But `render_prune.py` lives in the **plugin** (`plugins/agami/scripts/`), not the package — so that path is `site-packages/` (installed) or `src/` (dev), neither of which has it → `ModuleNotFoundError`. **Reproduced** in a clean `uvx --from "agami-core[model]"` env. It slipped through the run-throughs because the sample path skips discover and the real-DB attempt stalled earlier. Same OCR-028 package-split leftover class as the earlier install blockers.

## Changes

- **`cli.py`** — new `_load_render_prune()` helper that locates `render_prune` via **`AGAMI_PLUGIN_ROOT/scripts`** (set by every `sm` caller, inherited by the CLI subprocess; an explicit `sys.path.insert` survives `sm`'s `cd /` + `PYTHONSAFEPATH`), with a package-relative dev fallback and a **clear error** naming `AGAMI_PLUGIN_ROOT`. `cmd_discover` uses it. Two tests drive the locator directly (no DB needed — the DB-driven `cmd_discover` path the old test never exercised).
- **Hygiene:** removed the inert `--in-place` arg from `setup_desktop_mcp.py` (OCR-033 made it a no-op) + its three stale doc mentions; slimmed `test_skill_guardrails.py` to the one guardrail with no code equivalent; dropped the over-broad `"secrets." not in raw` assertion from `test_release_pypi_workflow.py`.

## Test plan

- `uvx --with-editable "packages/agami-core[model,server]" --with pytest pytest tests/` → **1038 passed**; ruff + gitleaks clean.
- `test_load_render_prune_via_plugin_root` (resolves via `AGAMI_PLUGIN_ROOT`) + `test_load_render_prune_raises_without_plugin_root` (clear error) lock the fix; the file-check gates a stale `sys.modules` so the raise is non-spurious.
- Real `sm discover` on a marketplace install (renders the prune page) is a manual check — needs a live DB.

## Checklist

- [x] `Spec: OCR-035`
- [x] Full gate green (1038 tests, ruff, gitleaks)
- [x] New behavior ships with tests; no test weakened without justification (the slimmed/loosened tests keep their real invariants — reviewer-confirmed)
- [x] No secrets / no customer data; no new spec-IDs added to code
- [x] Decision logged (locate via `AGAMI_PLUGIN_ROOT`, not vendor a presentation script into the pip package)

Spec: OCR-035

### DIFF
--- a/docs/mcp-server.md
+++ b/docs/mcp-server.md
@@ -101,9 +101,9 @@ It removes all three sharp edges of hand-editing:
   backup, atomic write, every other key and MCP server preserved.
 
 Then **fully quit** the app (Cmd+Q on macOS; quit from the system tray on Windows)
-and reopen. Flags: `--profile NAME` (pin a profile), `--in-place` (editable install
-from a checkout — for devs iterating on the server), `--config PATH`
-(target a different client's config file). Re-run after a plugin update to reinstall.
+and reopen. Flags: `--profile NAME` (pin a profile), `--config PATH`
+(target a different client's config file), `--python /abs/python3` (force an interpreter).
+Re-run after a plugin update to reinstall.
 The helper auto-resolves the macOS vs Windows config path for you.
 
 ### Manual (what the helper does under the hood)
--- a/packages/agami-core/src/semantic_model/cli.py
+++ b/packages/agami-core/src/semantic_model/cli.py
@@ -937,27 +937,24 @@ def _route_references(org, specs: list[dict]) -> tuple[dict, list, int]:
 def _load_render_prune():
     """Import the `render_prune` script (the prune-page renderer) and return the module.
 
-    render_prune lives in the PLUGIN (`plugins/agami/scripts/`), not this package, so a
-    package-relative import fails on any installed layout (`site-packages/` when pip-installed,
-    `src/` in a dev checkout — neither has it). Locate it via `AGAMI_PLUGIN_ROOT`, which every `sm`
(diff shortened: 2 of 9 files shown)


ashwin-agami:
All 3 addressed:

1. **Misleading dev fallback** — right, the package-relative candidate never held `render_prune.py`. Made it **env-only + honest**: `sm` now **exports `AGAMI_PLUGIN_ROOT`** (derived from its own location if the caller didn't set it), so the locator is guaranteed to resolve in the CLI subprocess; `_load_render_prune` drops the dead fallback and raises a clear error otherwise.
2. **Vacuous test** — the tests now strip `render_prune` from `sys.modules` + the scripts dir from `sys.path` before calling, so they genuinely exercise the locator (fail if it stops inserting the path).
3. **Lost guard** — re-added a **targeted** version: assert no `gh-action-pypi-publish` step references `secrets.` (scoped to the publish steps, so a legit `secrets.GITHUB_TOKEN` elsewhere can't false-trip).

PR #74, opened 2026-07-04
Env-var warehouse credentials in the shared executor (env-first, file-fallback)

ashwin-agami:
## Summary
`execute_sql._load_credentials` gains an **environment** source resolved *ahead of* the `local/credentials` file. This is the self-host / container credential channel: a mounted 600-mode creds file can't be read by a container running as a different uid (and `execute_sql` rejects the file if perms are widened) — env has no file mode and is inherited by the subprocess, so it sidesteps that whole class. Motivated by a field deploy where queries failed on `Permission denied` reading the mounted credentials.

It's an added **source, not a fork** (REQ-002): with no env var set, resolution is byte-for-byte the old file path — the local plugin and its chmod-600 gate are untouched. Env secrets are the cloud-neutral pattern (REQ-007).

## Changes
- `packages/agami-core/src/execute_sql.py`
  - New `_env_datasource_dsn(profile)` — returns a DSN from `DATASOURCE_URL__<PROFILE>` (per-datasource; profile id upper-cased, non-alphanumeric → `_`) then `DATASOURCE_URL` (single default), or `None`.
  - `_load_credentials` tries env first (parsed by the **existing** `_parse_dsn`, so the DSN scheme carries the warehouse type and query params merge), else the unchanged file read.
  - Missing-both error now names **both** `DATASOURCE_URL` and the file. Module docstring corrected.
- `tests/test_env_credentials.py` — 6 tests: env-first with no file, per-datasource precedence, profile-id normalization, snowflake scheme+params via env, no-env-falls-through-to-file (no-fork regression), missing-both error names both.
- `plugins/agami/lib/execute_sql.py` — re-synced (`dev.py sync-lib`).

## Checklist
- [x] `uv run dev.py check` green — **1044 passed**, ruff clean, gitleaks clean
- [x] New behavior covered by tests (no live DB — env parsing + precedence + no-file path)
- [x] No fork: file path + 600 gate unchanged when no env is set (regression test)
- [x] No secret on a command line / in logs; env inherited by the subprocess
- [x] Customer-safety: neutral placeholders only (`warehouse`, `acme`-style), no real names

Consumed by ACE-027 (deploy wiring: `.env` carries the DSN, bundle drops `local/`).

Spec: ACE-026

### DIFF
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -106,13 +106,19 @@ def _env_datasource_dsn(profile: str) -> str | None:
         env channel supports the schemes `_parse_dsn` handles (postgres / redshift /
         mysql / snowflake / bigquery / sqlite). A warehouse type without a DSN scheme
         (databricks, oracle, sqlserver, trino, duckdb) still uses the per-field file.
-      - An empty value is treated as unset (falls through to the next source) — set the
-        var to a real DSN to take effect; don't set it to "" expecting to *disable* one.
+      - The value is stripped (secret stores / `.env` / `$(cat file)` commonly append a
+        trailing newline, which would otherwise mis-parse), and an empty-or-whitespace-only
+        value is treated as unset (falls through to the next source) — set the var to a real
+        DSN to take effect; don't set it to "" expecting to *disable* one.
       - The token folds every non-alphanumeric char to `_`, so profiles differing only in
         punctuation (`sales-pg` vs `sales.pg`) map to the same var — name profiles distinctly.
     """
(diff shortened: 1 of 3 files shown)


ashwin-agami:
Addressed Copilot's review in e54001c: `_env_datasource_dsn` now strips env DSN values (matching the file path's per-field `.strip()`) so a trailing newline from a secret store / `.env` / `$(cat file)` parses cleanly, and treats an empty-or-whitespace-only value as unset (falls through to the next source). Added a regression test.

ashwin-agami:
Addressed: extracted `_env_token(profile)` and reuse it in both the DSN lookup and the missing-creds error message (was recomputed inline). No behavior change.

PR #76, opened 2026-07-04
Deploy: env-var warehouse creds + world-readable model (no mounted secret, no perm crash)

ashwin-agami:
## Summary
The `/agami-deploy` bundle mounted the operator's entire `~/agami-artifacts` (incl. the 600-mode `local/credentials`) read-only into a container that runs as **uid 10001**. Two field failures, both from the uid mismatch: (1) the model files weren't world-readable → boot-time `model_deploy` crashed `Permission denied: ORGANIZATION.md` → 502 crash-loop; (2) even past that, `execute_sql` couldn't read the 600 creds file (and rejects it if perms are widened).

This makes the deploy consume ACE-026's env-var creds: the warehouse DSN travels in `.env` (`DATASOURCE_URL`), the bundle **stops shipping `local/`**, and the (non-secret) model is staged **world-readable**. Net: no secret is ever mounted, and the operator never runs `chmod`.

## Changes
- `plugins/agami/scripts/prepare_deploy.py`
  - Stage the **model only** — `copytree(..., ignore=ignore_patterns("local"))` — so `local/credentials` and `local/.pgpass` never enter the bundle.
  - `_grant_world_read` widens the staged model (`chmod -R a+rX` semantics: add-only, dirs traversable + files readable; symlinks skipped) so uid 10001 reads it through the read-only mount.
- `deploy/.env.example` + `plugins/agami/skills/agami-deploy/bundle/.env.example` — delete the now-false NOTE ("credentials travel in your mounted artifacts") and add a commented `DATASOURCE_URL` hint (+ `DATASOURCE_URL__<NAME>` for a second warehouse). It's a credential, so the helper never writes it.
- `plugins/agami/skills/agami-deploy/SKILL.md` — Phase-2 hand-off asks the user to set `DATASOURCE_URL` by hand (alongside the admin password), and notes the warehouse creds are no longer shipped.
- `tests/test_prepare_deploy.py` — +4: `local/` excluded, `.pgpass` excluded, staged model world-readable (the crash regression), `DATASOURCE_URL` hint ships but is left unset. Updated the old "credentials staged" assertion to its opposite.

## Checklist
- [x] `uv run dev.py check` green — ruff clean, gitleaks clean, full suite passes
- [x] Crash regression covered: staged model dirs others-`r-x`, files others-`r` (repairs a 0o700/0o600 source)
- [x] No secret in a shippable bundle or a mounted volume; DSN is a `.env` hand-off, never on a command line
- [x] Customer-safety: neutral placeholders only (`acme`, `your-warehouse.example`, `you@example.com`)

Depends on **ACE-026** (#74) for the executor to read `DATASOURCE_URL`. Also revises the ACE-009 `.env` contract (creds file → env).

Spec: ACE-027

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -17,8 +17,12 @@ AGAMI_ADMIN_PASSWORD=choose-a-strong-password
 # AGAMI_OIDC_GOOGLE_CLIENT_ID=
 # AGAMI_OIDC_GOOGLE_CLIENT_SECRET=
 
-# NOTE: the WAREHOUSE the model queries is NOT configured here — its credentials travel in your mounted
-# artifacts at `<artifacts>/local/credentials` (an INI keyed by datasource), the same file agami uses locally.
+# The WAREHOUSE the model queries — its connection DSN. This is a CREDENTIAL: type it here by hand and
+# never share it in chat. The scheme picks the type (postgresql:// mysql:// redshift:// snowflake://…);
+# add query params as needed (e.g. ?sslmode=require, or Snowflake ?warehouse=…&role=…). For a SECOND
+# datasource, add DATASOURCE_URL__<NAME> (e.g. DATASOURCE_URL__SALES=…) — <NAME> is the datasource id
+# upper-cased with non-alphanumerics folded to `_`.
+# DATASOURCE_URL=postgresql://user:pass@your-warehouse.example:5432/analytics?sslmode=require
 
 # External/managed Postgres for the app's own state. Omit to use the bundled `postgres` (bundled-db profile).
 # Use a PLAIN postgresql:// URL (e.g. a Cloud-SQL-like service via its IP + sslmode=require) — not a cloud connector.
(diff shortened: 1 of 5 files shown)


ashwin-agami:
Addressed Copilot's review:
- **rglob symlink traversal (prepare_deploy.py):** switched `_grant_world_read` to `os.walk(followlinks=False)` — `rglob('**')` follows directory symlinks on Python ≤3.12 (our CI runs 3.10–3.12), which could chmod files outside the staged bundle. `os.walk` never enters a symlinked dir and streams (no giant path list). Added a regression test proving an external dir reached via a symlink in the model is left untouched.
- **test start-of-file robustness:** the DATASOURCE_URL 'never written live' assert is now a per-line check, so it holds even if the assignment were at byte 0 (and doesn't false-match `DATASOURCE_URL__<NAME>`).

ashwin-agami:
Addressed Copilot's re-review:
- **Stale `artifacts/local` on re-run (real bug):** `copytree(ignore='local')` skips copying but doesn't delete a `local/` staged by an *older* prepare_deploy — so re-running over an old bundle kept a stale credentials file. Now `rmtree` the staged `local/` before copy. Added a regression test.
- **`_widen_one` docstring:** fixed the contradictory 'never wider' wording (it widens; meant 'never narrower').

ashwin-agami:
Addressed: the bundle `.env.example` header no longer says the password is the *only* hand-typed value — it now names DATASOURCE_URL (and APP_DATABASE_URL / OIDC secrets when used) as credentials you fill by hand.

PR #78, opened 2026-07-04
Deploy config is a visible agami.env at a confirmed location

ashwin-agami:
## Summary
The one file a self-hoster must edit was written as **`.env`** — a dot-file, **hidden in Finder** on macOS. This renames it to **`agami.env`** (visible) across the bundle, and makes the skill **confirm the target directory** and **open the file** for the user so they never hunt for it.

Docker-compose only auto-loads a file literally named `.env`, so everything now references the file explicitly via `env_file:` / `--env-file agami.env`.

## Changes
- `prepare_deploy.py` — writes `agami.env` (from renamed `agami.env.example`), still chmod 600; status/docstrings updated.
- `bundle/` + `deploy/` `docker-compose.yml` — `env_file: agami.env`; header points at `./deploy.sh`.
- `bundle/deploy.sh` (`pull`/`up`) + `deploy/deploy.sh` (`preflight`/`up --build`) — pass `--env-file agami.env`.
- `deploy_preflight.py` — default + messages target `agami.env`.
- `SKILL.md` — **Phase 1** confirms the target dir (never inside artifacts); **Phase 2** `open -t <target>/agami.env` + prints the absolute path.
- `deploy/.gitignore`, both READMEs, both `.env.example` → `agami.env.example` (git mv).
- Tests updated to `agami.env`; new `test_config_is_visible_agami_env_wired_through_compose_and_deploy`.

## Checklist
- [x] `uv run dev.py check` green — ruff clean, gitleaks clean, full suite passes
- [x] Test pins: `agami.env` exists, no `.env`, compose `env_file:` + `deploy.sh --env-file` wire it through
- [x] Still chmod 600; no secret on a command line; no behavior change beyond the filename + skill conversation

First of two deploy-UX slices. **ACE-031** (upgrade-aware re-run + multi-datasource) builds on this.

Spec: ACE-030


ashwin-agami:
Agami review found two **must-fix** misses in the manual `deploy/` (git-clone) path — the staged-bundle tests didn't cover it. Both fixed in dd1c313:
- **deploy/README.md variants table** ran bare `docker compose up -d`; since `COMPOSE_PROFILES` lives in `agami.env` (which compose no longer auto-loads), that would start only `agami` and skip Caddy/Postgres. Now `docker compose --env-file agami.env up -d`.
- **deploy/agami.env.example** header said *Copy to `.env`* → now `agami.env`.

Also swept config-sense `.env` comment refs → `agami.env`, and fixed a stale `deploy_preflight` comment that still claimed warehouse creds travel in the mounted artifacts (pre-ACE-027 — they're `DATASOURCE_URL` now).

ashwin-agami:
**GitGuardian alert = false positive, now hardened.** The flagged `AGAMI_ADMIN_PASSWORD=choose-a-strong-password` in `deploy/agami.env.example` is a placeholder (no real secret; the `git mv` rename made GitGuardian re-scan it as a new file). But a *non-empty* placeholder also passes preflight's 'password is set' check, so I blanked it (matching the bundle template) — a genuine footgun avoided. Also angle-bracketed the DSN placeholder (`<user>:<password>`) in both examples.

Copilot's two comments addressed:
- `deploy_preflight` missing-file error no longer assumes `agami.env.example` is present (a bundle ships only the generated `agami.env`).
- `open -t --` guards a target path that starts with `-`.

PR #79, opened 2026-07-04
Upgrade-aware /agami-deploy re-run + multi-datasource select

ashwin-agami:
## Summary
A re-run of `/agami-deploy` used to leave an existing `agami.env` **completely untouched** — safe for secrets, but it meant a version that added a required setting (like v0.3.5's `DATASOURCE_URL`) never surfaced, and a pinned image never moved. And multiple models deployed silently. This makes the re-run **upgrade-aware** and multi-datasource an **explicit choice** — an upgrade now feels like the first deploy: run the skill, fill anything genuinely new, done.

## Changes
- `prepare_deploy.py`
  - **`_merge_env`** (non-destructive upgrade): keeps every existing value (never touches a typed password/secret), **appends any template key new in this version** (as its commented hint — so `DATASOURCE_URL` shows up), and **bumps `AGAMI_IMAGE_TAG` only if `--image-tag` is passed** (a model-only re-stage leaves the pin). New status **`UPGRADED <dir> new_keys=<a,b,…>`**.
  - **`--datasources a,b`** — selective staging via a `copytree` ignore that drops non-chosen profile dirs, always keeping `local/`-exclusion + install-global files (default: all models).
  - `--image-tag` now defaults to `None` (fresh → `latest`).
- `SKILL.md`
  - Phase 1 asks **which datasource(s)** to deploy when there's more than one; reads `PREPARED` vs `UPGRADED`+`new_keys` and surfaces the new settings to the user.
  - Phase 2 gives **per-datasource DSN guidance** (`DATASOURCE_URL` / `DATASOURCE_URL__<TOKEN>`).
  - "Re-running later" splits **model update** (no `--image-tag`, restart) vs **version upgrade** (`--image-tag`, `./deploy.sh`).
- `tests` — +7 (merge preserves secrets & surfaces new keys, tag-bump-only-when-passed, no-spurious-keys, selective staging); re-run test updated to `UPGRADED`.

## Checklist
- [x] Merge invariant tested: existing values byte-preserved; `new_keys` accurate; tag bumps only when passed
- [x] Secret-safe: no existing value changed/removed; no secret logged or on a command line
- [x] Placeholders are `<user>:<password>` (no GitGuardian-tripping `user:pass`)

Builds on **ACE-030** (visible `agami.env`) and **ACE-026** (`DATASOURCE_URL`). Completes the deploy-UX pair.

Spec: ACE-031

### DIFF
--- a/plugins/agami/scripts/prepare_deploy.py
+++ b/plugins/agami/scripts/prepare_deploy.py
@@ -106,10 +106,12 @@ def _env_key(line: str) -> str | None:
 
 
 def _merge_env(existing: str, template: str, image_tag: str | None) -> tuple[str, list[str]]:
-    """Non-destructive upgrade of an existing `agami.env`: **keep every existing line/value** (never touch a
+    """Non-destructive upgrade of an existing `agami.env`: **preserve every existing value** (never touch a
     typed password / generated secret), **append any template key not already present** (as the template's
     own line — a commented hint or a value, so a key new in this version like `DATASOURCE_URL` shows up), and
-    — only when `image_tag` is given — bump the non-secret `AGAMI_IMAGE_TAG`. Returns (merged_text, new_keys)."""
+    — only when `image_tag` is given — bump the non-secret `AGAMI_IMAGE_TAG`. The output is LF-normalized
+    (docker's `env_file` + the generated file use LF; a value's content is unchanged by that). Returns
+    (merged_text, new_keys)."""
     present = {k for line in existing.splitlines() if (k := _env_key(line))}
     merged = existing if existing.endswith("\n") else existing + "\n"
     new_lines: list[str] = []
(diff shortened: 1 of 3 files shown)


ashwin-agami:
Addressed all three Copilot comments:
- **_env_key strip order (real bug):** stripped `#` before whitespace, so an *indented* commented key (`  # KEY=…`) wasn't recognized → `_merge_env` could re-append it as a duplicate. Now strips whitespace first (+regression test).
- **--datasources all-unknown:** now fails fast with `ERROR … matched no model` instead of building a modelless bundle; a *partial* unknown still warns + stages the valid ones (+test).
- **_args() default:** now `image_tag=None` (matches the CLI), so tests exercise the real 'don't bump unless passed' default.

ashwin-agami:
Addressed both:
- **Stale datasource on re-run:** `--datasources` now purges previously-staged model dirs not in the chosen set (copytree merges, won't delete — so a dropped datasource would otherwise linger and still be served). +regression test.
- **CRLF:** `_merge_env` LF-normalizes its output, so appending LF lines to a CRLF file can't mix newlines. Values are unchanged; reworded the invariant to 'value-preserved' (line endings normalize to LF, which is docker's env_file format). +test.

ashwin-agami:
Addressed: both purge paths (datasource-drop + stale-local) now **unlink a symlink** instead of `rmtree`-ing it (rmtree raises on a symlink, and copytree(symlinks=True) can stage a symlinked model) — the link is removed, its target (possibly outside the bundle) is never touched. +regression test.

PR #82, opened 2026-07-05
docs(agami): nudge users to connect with a read-only DB user

sandeep-agami:
## What

agami only ever runs **read-only SELECT** queries (query generation refuses DML/DDL; the local MCP server double-checks every statement). So the right thing to connect it to is a **read-only database user**. Today nothing in setup says that or shows how — a few example usernames say `readonly` but there's no recommendation and no grant SQL.

This adds a **non-blocking** read-only recommendation at every credential-collection point, made **actionable** with copy-paste `GRANT` SQL. Read-write credentials still work; nothing is gated.

## Changes

- **New [`plugins/agami/shared/readonly-grants.md`](../blob/readonly-user-nudge/plugins/agami/shared/readonly-grants.md)** — copy-paste `CREATE USER` / `GRANT SELECT` per dialect: Postgres/Redshift, MySQL, Snowflake, SQL Server, Oracle, Databricks (Unity Catalog), Trino, BigQuery (IAM), plus a SQLite/DuckDB file-permission note.
- **`agami-connect` Phase 0a** — read-only lines in the `credentials.example` header (lands in every user's file) + a hand-off **Tip** shown only for dialects with a user/role concept (skipped for sample / SQLite / DuckDB / BigQuery-ADC).
- **`agami-deploy`** — nudge in the Phase 2 `DATASOURCE_URL` hand-off and in **both** `agami.env.example` templates (skill `bundle/` + top-level `deploy/`), plus a **Security notes** bullet in `deploy/README.md`.
- **Reference / public docs** — `credentials-format.md`, `docs/credentials.md`, `docs/usage.md`, `README.md`, and a `CHANGELOG` (Unreleased) entry.

Every surface links to the new grants doc (relative links verified to resolve).

## Scope notes

- No changes to the connection / introspection code path — guidance + docs only.
- Deliberately skipped `SECURITY.md` (a vuln-reporting policy) and `docs/trust-layer.md` (model confidence/review) as not natural fits; `docs/self-hosting.md` documents the app **store** DSN, not the warehouse datasource, so no read-only row there.
- Possible follow-up (out of scope here): an active privilege probe at introspect time that warns if the connected user can write.

## Verification

- All relative links to `readonly-grants.md` resolve from each file.
- Dry-run the flow: `/agami-connect` → PostgreSQL shows the nudge in the template + hand-off; the sample-database path shows none.
- Ask "give me the read-only grant" and confirm the Postgres block comes back with `<db>`/`<schema>`/`<user>` filled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ is the source of truth a host installs against — bumping it is what invalidate
 user's plugin cache (see [CONTRIBUTING.md](CONTRIBUTING.md)). Each released section
 below corresponds to one such version.
 
+## [Unreleased]
+
+### Added
+
+- **Read-only database user guidance.** `/agami-connect` and `/agami-deploy` now
+  recommend connecting agami with a **read-only** database user — agami only ever
+  runs read-only SELECT queries, so read access is all it needs. A new
+  [readonly-grants.md](plugins/agami/shared/readonly-grants.md) ships copy-paste
+  `CREATE USER` / `GRANT SELECT` SQL for every supported dialect (Postgres/Redshift,
+  MySQL, Snowflake, SQL Server, Oracle, Databricks, Trino, BigQuery). Ask agami for
+  "the read-only grant" to get the exact SQL for your database.
+
 ## [0.3.6] — 2026-07-04
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ When you're ready to connect your own database:
 /agami-connect
 
 # 2. Fill in the template, then save + lock it down
+#    (agami only runs read-only queries, so use a read-only DB user if you can —
+#     ask agami for "the read-only grant" to get the exact GRANT SQL)
 $EDITOR <artifacts_dir>/local/credentials.example
 mv <artifacts_dir>/local/credentials.example <artifacts_dir>/local/credentials
 chmod 600 <artifacts_dir>/local/credentials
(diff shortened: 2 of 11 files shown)


sandeep-agami:
Thanks for the review — all six addressed in c1d785d:

1. **SQL Server / Azure SQL (`CREATE LOGIN`)** — split into **SQL Server / Azure SQL Managed Instance** (login + mapped user) and **Azure SQL Database** (contained user, `CREATE USER … WITH PASSWORD`, no login), since Azure SQL Database doesn't support server logins.
2. **Incomplete placeholder list** — expanded the intro to list every placeholder used across blocks (`<project>`, `<dataset>`, the `agami_ro` name, etc.) and noted each block uses only some.
3. **SQLite "read-only mode" claim** — verified against `plugins/agami/lib/execute_sql.py`: SQLite uses `sqlite3.connect(path)` (not read-only), only DuckDB uses `read_only=True`. Rewrote the note so safety is credited to the read-only SQL guard + filesystem permissions, with the read-only-copy suggestion for SQLite.
4. **Assistant fill-in too narrow** (`agami-connect`) — generalized from `<db>/<schema>/<user>` to "whatever `<…>` placeholders the block uses (e.g. `<warehouse>`/`<catalog>`/`<project>`/`<dataset>`)".
5. **`deploy/agami.env.example` path** — the repo-relative path won't resolve after `cd deploy`; switched to the stable GitHub URL (`…/blob/main/plugins/agami/shared/readonly-grants.md`).
6. **Self-contained bundle env** — same fix: the bundle is copied to a host with no repo, so it now points to the stable GitHub URL and "ask agami for the read-only grant" rather than a local path.

The in-repo doc/skill cross-links (markdown, viewed in the repo) keep their relative paths; only the two `agami.env` templates — which land on a host or a `cd deploy` cwd — use the absolute URL.

PR #100, opened 2026-07-10
feat(mcp): tool-extension + adapter-injection seam for the HTTP server

vishalkalbi27:
## Summary

Adds a supported composition seam so a downstream consumer can extend the MCP server **without forking or monkeypatching core**. Additive and **no-op by default** — an existing deploy behaves exactly as before.

## Changes

- **`tools.register(name, handler, description, inputSchema)`** — adds a tool to the shared `TOOLS` registry, with a **duplicate-name guard** so a consumer can't silently shadow a core tool like `execute_sql`.
- **`mcp_http.create_app(extra_tools={}, adapters=None)`** — a composition factory that merges `extra_tools` over a **copy** of `TOOLS` (never mutating the module global) and injects the four `ports.py` adapters, defaulting to the OSS adapters when `None`.
- **`ports.Adapters`** — a frozen dataclass bundling the four port adapters so they pass as one argument.
- **`build_app()`** is retained as a thin `create_app()` wrapper.

## Backwards compatibility

- `create_app()` with no args behaves **identically** to the previous `build_app()` — same routes, same `tools/list`, same auth challenge.
- `execute_sql`'s `inputSchema` stays **byte-identical**.
- The module-global `TOOLS` is never mutated by `create_app`.
- `python -m mcp_http` / `main()` and the stdio entrypoint (`mcp_harness`) are unchanged.

## Tests

`tests/test_tool_extension_seam.py` covers: no-arg parity with `build_app`, byte-identical `execute_sql` schema, the non-mutating registry merge, the duplicate-name guard, OSS-default adapters vs. a passed `Adapters(...)`, and the unchanged auth challenge. `ruff` clean; the new + existing MCP/tools/harness suites pass.

### DIFF
--- a/packages/agami-core/src/mcp_http.py
+++ b/packages/agami-core/src/mcp_http.py
@@ -27,8 +27,13 @@ from pathlib import Path
 import admin
 import onboarding
 import user_store
-from oss_adapters import PresenceAuthProvider, SingleTenantOrgResolver
-from ports import AuthProvider, Org
+from oss_adapters import (
+    FileActivitySink,
+    PresenceAuthProvider,
+    SingleTenantOrgResolver,
+    WarnOnlyGovernancePolicy,
+)
+from ports import Adapters, AuthProvider, Org
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.middleware.base import BaseHTTPMiddleware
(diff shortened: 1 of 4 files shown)


vishalkalbi27:
I have read the CLA Document and I hereby sign the CLA

vishalkalbi27:
recheck

PR #101, opened 2026-07-10
feat(store): migration-overlay seam — layer extra migration roots on core

vishalkalbi27:
## Summary

Lets the migration runner apply **additional migration roots** on top of `migrations/core`, in deterministic order, so a downstream consumer can add its own tables **without editing core's migration tree**. Additive and **no-op by default**.

## Changes

- **`Store.run_migrations(migrations_dir=None, overlay_dirs=None)`** — applies core first, then each overlay root in order.
- **`register_migration_overlay(path)`** — registers an overlay root so the no-arg `run_migrations()` call in the `mcp_http` lifespan picks it up, with **no call-site edit**.
- Overlay tracking ids are **namespaced by the root's directory name**, so a core and an overlay file with the same filename can't collide on the `schema_migrations` primary key. Core ids stay bare filenames.

## Backwards compatibility

- With no overlays, a run applies **exactly** `migrations/core` — already-migrated DBs are **byte-identical** (core ids unchanged).
- Both existing call sites (`mcp_http` lifespan, `model_deploy`) are unchanged no-arg calls.
- The Postgres advisory-lock bracket, fail-closed rollback, and per-migration commit are preserved.

## Correctness

Idempotent (a second run is a no-op), **order-deterministic** (core → overlay-1 → overlay-2 …), collision-safe via namespacing, and resume-safe (overlays recorded in `schema_migrations` like core migrations).

## Tests

`tests/test_auto_migrate.py` adds: bare-core-ids on the no-overlay path, core-then-overlay + idempotency, deterministic ordering, same-filename collision avoidance, resume after a later-added overlay, and the registered-overlay no-arg path. `ruff` clean; the migration suite passes in isolation.

### DIFF
--- a/packages/agami-core/src/store.py
+++ b/packages/agami-core/src/store.py
@@ -38,6 +38,21 @@ def _now_iso() -> str:
     return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
 
 
+# Extra migration roots layered on top of migrations/core. A consumer registers its root here at
+# boot so the mcp_http lifespan's `run_migrations()` (no args) applies it — no call-site edit.
+_MIGRATION_OVERLAYS: list[Path] = []
+
+
+def register_migration_overlay(path: Path) -> None:
+    """Register an extra migration root, applied AFTER migrations/core in registration order.
+
+    Lets a consumer overlay its own tables without editing core's tree. Roots should have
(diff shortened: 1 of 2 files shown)


vishalkalbi27:
I have read the CLA Document and I hereby sign the CLA

vishalkalbi27:
recheck

PR #103, opened 2026-07-11
ACE-045: per-process model cache + single SQL parse + shared guard indices

ashwin-agami:
## Summary
Behaviour-preserving perf hardening (audit finding **P2**, feature **F12-runtime-scalability**). The semantic model is parsed/indexed far more than needed per query. Two independent wins, same answers:

- **Slice 1 — GuardContext (both paths).** The `_model_safety` guard battery re-parsed the SQL **6×** and rebuilt each model index per guard. A `GuardContext` now parses once + builds each index once, threaded through the guards.
- **Slice 2 — get_cached_org (long-lived server).** The parent loaded the whole model **2–3× per query** and re-loaded it every query. It's now cached per process, keyed `(org_id, datasource, model_version)`, served warm across queries + users.

## Changes
- `semantic_model/runtime.py` — `GuardContext` + `build_guard_context`; optional `ctx=` on the 6 guards + `_preflight_select`/`_check_aggregation_semantics`. `ctx=None` keeps the standalone (`cli.py`) path **byte-identical**.
- `execute_sql.py` — build the context once in `_model_safety`, thread it; rebuild after an auto-rewrite (SQL changes).
- `tools.py` — `get_cached_org` + `_current_org_ctx`/`_current_org_id`; route `_resolve_units`/`_resolve_receipt`/`tool_get_datasource_schema`. **Bypass the cache when `model_version` is None** (never serve a stale model).
- `mcp_http.py` — set `_current_org_ctx` per request in `handle_mcp` from the resolved org (mirrors `_actor_ctx`). **Tenant-safe by construction**: the cache key is org-scoped via the `OrgResolver` seam — N=1 today, per-request under a multi-tenant resolver, **no fork, no F14 dependency**.
- `tests/` — `test_guard_context.py`, `test_org_cache.py`, `conftest.py` (isolates the module-global cache).
- Synced the bundled `lib/` mirror.

## Test plan / acceptance
- **Slice 1:** parse=1 + each-index-built=1 counter; verdict-parity with/without `ctx` across 6 SQL cases; unparseable→allow.
- **Slice 2:** 1 load/query + 0 on repeat; reload after a version bump; two orgs on the same datasource get **separate** models (no cross-tenant); default org_id → `local`.
- **Full gate green:** 1413+ tests, ruff + format + gitleaks + lib-drift all pass (`uv run dev.py check`).

## Out of scope (per spec)
Cross-query caching on the subprocess/file path (needs ACE-028); the multi-tenant resolver (agami-hosted F5) — this only *consumes* the `OrgResolver` port.

Spec: ACE-045

### DIFF
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -736,10 +736,16 @@ def _model_safety(sql: str, profile: str, area: str | None):
         sys.stderr.write(f"[agami] could not load semantic model; skipping safety pass: {e}\n")
         return sql, None
 
+    # Build the shared guard context ONCE — parse the SQL + build each model index a single
+    # time — and thread it through the battery below, instead of every guard re-parsing and
+    # rebuilding its index (audit P2 / ACE-045). Behaviour-preserving: a guard given `ctx`
+    # returns the same verdict as one that builds its own.
+    ctx = RT.build_guard_context(sql, org)
+
     # Table-scope guard — a query may only reference tables the semantic model
     # declares; any other table in the connected database is refused. Runs FIRST
     # so the fan/chasm and sensitive checks below only evaluate in-scope tables.
-    ts = RT.check_table_scope(sql, org)
+    ts = RT.check_table_scope(sql, org, ctx=ctx)
     if ts.action == "refuse":
         json.dump({"error": {"kind": "table_out_of_scope", "tables": ts.offending_tables,
                              "reason": ts.reason, "suggestion": ts.suggestion}}, sys.stderr)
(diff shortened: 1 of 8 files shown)


ashwin-agami:
Thanks @copilot — all three addressed:

- **build_guard_context builds indices even when sqlglot is unavailable** — good catch; it now returns \`None\` early when \`not _HAVE_SQLGLOT\`, since every guard short-circuits to allow before it touches the context. Fixed centrally in \`build_guard_context\` so both call sites (initial + post-rewrite rebuild) benefit.
- **pre_flight_check parity for unparseable SQL** — real divergence in the byte-identical \`.as_dict()\` invariant this spec claims: with \`ctx\`, \`ctx.tree\` is \`None\` → "no SELECT; skipped", but the non-ctx path raised → "unparseable; skipped (…)". Unified by routing the non-ctx branch through the same \`_parse_sql\` helper, so both paths are now identical.
- **duplicate autouse cache fixture** — removed the per-file \`_isolate_cache\`; \`conftest._reset_org_cache\` already covers it (and dropped the now-unused \`pytest\` import).

Gate green (1417 passed). Pushed.

PR #111, opened 2026-07-12
ACE-035: shared guardrail contract — Verdict / policy / Envelope

sandeep-agami:
## Summary

Introduces one shared result shape for the SQL-execution guardrails and refactors the safety gates + the `execute_sql` result wire onto it. Every gate now produces a `Verdict`; one `policy(verdict, tier)` maps a verdict to an action by class + deployment tier; and every surface (stdio + HTTP MCP) returns one `Envelope`. The ad-hoc gate return shapes and the hand-rolled error dicts are deleted, not wrapped. This is the root of the safe-SQL-execution work — the data-protection and governance gates build to these same types in their own changes.

## Changes

- **New `guardrail.py`** (stdlib-only, vendored): `Verdict` (a gate's finding), `Refusal`, `Envelope` (the one result shape), and `policy(verdict, tier) -> action`. The safety branch is fully implemented (reject in every tier; fail-closed on uncertainty); the data-protection and governance branches are documented stubs for later work.
- **Safety gates return `Verdict`**: `sql_guard.check_read_only` + `runtime.{check_table_scope, check_no_select_star, check_column_scope}` now return `Verdict | None` (None = allow); the three ad-hoc `*Result` dataclasses are deleted.
- **One `Envelope` on both surfaces**: the executor emits a `{"refusal": …}` line; `tools.tool_execute_sql` assembles the `Envelope` — a safety violation → `status=refused` + `refusal{kind,reason,remediation}` + no data; a clean query → `status=ok` + `data` (the existing `ExecuteSqlResult` payload) + `audit_id`. `ErrorResult`/`ToolError` folded into `Refusal`.
- **Audit trail**: new portable `guardrail_audit` table (migration 012); every ok/refused result writes one row keyed by `audit_id` at the shared chokepoint (so both surfaces are covered), best-effort.
- **Governance seam**: the old `GovernanceVerdict` result type is retired; the `GovernancePolicy` injection seam is kept, re-typed to return `list[Verdict]` (a no-op default for later governance work).

## Design decisions

- Kept the governance plug-point (re-typed to `Verdict`) rather than deleting it — later governance work injects through it.
- `Envelope` is the single top-level result; the well-modeled `ExecuteSqlResult` is reused as its `data` payload (not flattened). Pre-release, so no compatibility shims.
- `Refusal.kind` is an open `str` documented by `REFUSAL_KINDS` — operational failures (`syntax`/`auth`/`driver_missing`) come from the DB driver, not a fixed guardrail vocabulary — with a test pinning the emit sites against the documented set.
- The audit trail is a dedicated table written at the executor chokepoint: the generic tool-call log is written only by the HTTP transport, so it would miss the stdio surface.

## Test plan

- Full gate green: `uv run dev.py check` — **1494 tests**, ruff, format, gitleaks, and vendored-lib drift.
- New tests: `test_guardrail.py` (types + policy), `test_guardrail_audit.py` (audit trail), `test_execute_sql_envelope.py` (the executor→tool refusal relay), and `tests/e2e/test_safety_envelope.py` (one Envelope over the real stdio + HTTP transports).
- ~78 existing gate/wire assertions migrated to the new shape without weakening (refusals still assert refusal; offending names still checked).

## Checklist

- [x] Full gate green (lint, format, tests, secret scan, vendored-lib drift)
- [x] Both surfaces (stdio + HTTP) return the `Envelope` and are tested on both
- [x] No second SQL parser; the single `execute_sql` enforcement point is preserved
- [x] Adversarial safety tests pass; fail-closed behavior preserved
- [x] Reviewed (code-review + security-review panel); findings dispositioned (raw operational error-text sanitization is a separate change)

Spec: ACE-035

### DIFF
--- a/dev.py
+++ b/dev.py
@@ -42,7 +42,7 @@ _LIB_DST = _ROOT / "plugins" / "agami" / "lib"
 # The module-load import closure of the 4 scripts (all stdlib-only). tests/test_plugin_lib_resolution.py
 # fails if a script gains a module-level import that's missing here; a new *lazy* import from the package
 # would also need adding to this list.
-_VENDORED = ["agami_paths.py", "execute_sql.py", "sql_guard.py", "semantic_model/__init__.py", "semantic_model/units.py"]
+_VENDORED = ["agami_paths.py", "execute_sql.py", "sql_guard.py", "guardrail.py", "semantic_model/__init__.py", "semantic_model/units.py"]
 
 
 def run(cmd: list[str], *, allow_fail: bool = False) -> int:
--- a/migrations/core/012_guardrail_audit.sql
+++ b/migrations/core/012_guardrail_audit.sql
@@ -1,0 +1,23 @@
+-- The **guardrail audit trail** — one row per execute_sql call, keyed by the response Envelope's
+-- `audit_id`. Written best-effort at the shared executor chokepoint (`tools.tool_execute_sql`) so
+-- every ok/refused result is recorded on BOTH surfaces (stdio + HTTP), not only the HTTP path that
+-- writes `tool_calls`. This is the trail the Envelope's `audit_id` points at.
+--
+-- Deliberately thin for now — status + refusal kind + the query context; a richer verdict/action
+-- list is a follow-on. Portable CREATE TABLE (runs on SQLite + Postgres unchanged).
+
+CREATE TABLE guardrail_audit (
+    audit_id       TEXT PRIMARY KEY,        -- == the response Envelope's audit_id
+    ts             TEXT NOT NULL,           -- UTC ISO8601
+    datasource     TEXT,
+    status         TEXT NOT NULL,           -- 'ok' | 'refused'
+    refusal_kind   TEXT,                    -- the refusal kind when status = 'refused'
+    sql            TEXT,                    -- the query the caller sent (may be NULL on a bad request)
+    row_count      INTEGER,                 -- on an ok result
+    execution_ms   INTEGER,
+    correlation_id TEXT,                    -- the turn (one user question), self-reported; may be NULL
+    source         TEXT                     -- 'mcp_server'
+);
+
+-- The admin views read newest-first; index the time access path.
+CREATE INDEX idx_guardrail_audit_ts ON guardrail_audit (ts);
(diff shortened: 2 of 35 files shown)


sandeep-agami:
Addressed the Copilot review (all four were docstring-accuracy comments, no behavior change) in the latest commit:

1. **`ports.py` `GovernancePolicy`** — removed the "warn-only, *and* returns no findings" contradiction. The OSS default is a no-op (empty list); the warn/recommend/enforce tier posture is applied by `policy()`, not the adapter.
2. **`oss_adapters.py` `WarnOnlyGovernancePolicy`** — clarified that the name is the OSS *posture* (governance only ever warns, never enforces) while the default adapter, having no rules wired, emits no findings. I kept the name rather than renaming to `NoopGovernancePolicy`: it's pre-existing and its empty-warnings behavior is unchanged by this PR (the old `GovernanceVerdict(allowed=True, warnings=[])` was also empty), and it's referenced across `mcp_http` + tests. Happy to rename in a follow-up if you'd prefer.
3 & 4. **`Refusal` docstring (core + vendored copy)** — now distinguishes a **guardrail** refusal (clean `reason`, never raw SQL/DB text) from an **operational** failure (carries the driver's error text until sanitization lands separately), matching `_refusal_from_stderr`.

Full gate green after the change.

ashwin-agami (approved):

sandeep-agami:
Superseded by #170, merged as `9d3fd8d`.

This PR implements the pre-principles contract — `Verdict{class, rule, severity, certainty}`, `policy(verdict, tier) → Action` and a six-value `Action` enum. The governance principles removed the ground each of those stood on: nothing is tiered so `tier` has no job, there are two outcomes so `Action` is not an enum, `data_protection` went with principle 5, and `severity`/`certainty` existed only to feed the graded policy. `policy()` is deleted outright.

What landed instead is the reduced contract: `Refusal{reason, rule, detail, remediation}` with `reason` closed to principle 4's three values, and `Envelope{status, data, refusal, failure, receipt, audit_id}` with present-iff enforced at construction. Per the remediation plan §4, this branch is retired rather than merged — it was kept open as the reference diff for the re-author, which is now done.

Closed unmerged, deliberately. Recorded in ACE-035's `landed:` note.

PR #112, opened 2026-07-12
ACE-037: fail-closed on unscopable SQL

sandeep-agami:
> **Stacked on #111** (the shared guardrail contract) — it depends on that PR's `Verdict`/`Envelope`/`safety_verdict` and the refactored gates. Base is the `ACE-035-shared-guardrail-contract` branch; retarget to `main` once #111 merges. Review/merge #111 first.

## Summary

Closes a silent fail-*open* in the SQL object-scope gates: they only reject the `exp.Table` sources they *find*, so a query that passes the read-only guard but whose `FROM`/`JOIN` source can't be scoped — a table-function / `ROWS FROM`, a `VALUES` / `UNNEST` / `LATERAL` source, or SQL that doesn't parse at all — leaves the scope walk with nothing to reject and is allowed to run. This adds a fail-**closed** scopability gate: such a query is refused (`unscopable_sql`) rather than run blind, at the single shared executor, on every surface.

## Changes

- **New `runtime.check_scopable(sql, org, ctx) -> Verdict | None`** — refuses when the tree is absent/unparseable, has no SELECT, or any source isn't a scopable named `exp.Table` / derived `exp.Subquery`. Two whole-tree sweeps (an empty-name `exp.Table` catches table-functions / `ROWS FROM`; a `LATERAL` sweep; and a non-`Table`/`Subquery` `From`/`Join` source catches `VALUES`/`UNNEST`) — so every set-operation arm is covered. Reuses the existing single parse (`ctx.tree`) — **no second parser**. Inert when the model declares no tables (consistent with the object-scope gates).
- **Wired first in `execute_sql._model_safety`** (before the object-scope gates), so an unscopable query fails closed instead of reaching their degrade-to-allow branches.
- **`AGAMI_SQL_UNSCOPABLE_POSTURE` flag** — `enforce` (default) refuses; `warn` logs + allows (a staged-rollout escape hatch, never the shipped default; any non-`warn` value fails closed). Documented in both `agami.env.example` copies.

## Design decisions

- The gate is a separate, self-contained check placed first — it doesn't change each object-scope gate's internals; it closes the gap *around* them.
- Fail-closed is uniform: unparseable, no-SELECT, and parser-unavailable all refuse under `enforce`.
- Inert when no tables are declared — a deployment with no declared surface isn't scoping (the object-scope gates behave the same); it adds no new exposure.
- The gate never touches the datasource (it reads the parsed tree + the resolved model), so it produces the same verdict whether the model is file-served or DB-served — proven by extending the existing model-source parity test.

## Test plan

- Full gate green: `uv run dev.py check` — ruff, format, gitleaks, vendored-lib drift, and the pytest suite across py3.10–3.12.
- New `tests/test_scopable_gate.py` (unit corpus: named tables / CTE / derived subquery / comma-join / set-op / `SELECT 1` allow; table-functions / `ROWS FROM` / `VALUES` / `UNNEST` / `LATERAL` / `LATERAL VIEW` / unparseable / no-SELECT / parser-unavailable refuse; ctx-parity).
- `tests/test_ace051_fail_closed.py` additions: the differential corpus refuses through the *wired* `_model_safety` under `enforce`; `warn` logs + allows; any non-`warn` posture value fails closed; and the model-source parity test asserts an unscopable query yields the **same refusal** on a file-served vs DB-served model.
- Verified end-to-end on the real executor subprocess: `enforce` → a `{"refusal":{"kind":"unscopable_sql"}}` line + non-zero exit *before any DB connection*; `warn` → the log line, then proceeds.

## Checklist

- [x] Full gate green (lint, format, tests, secret scan, vendored-lib drift)
- [x] Fail-closed by default (`enforce`); `warn` cannot silently be the default; both surfaces covered
- [x] No second SQL parser (reuses the single parse); single `execute_sql` enforcement point preserved
- [x] Path-agnostic — same fail-closed verdict on file-served vs DB-served models
- [x] Reviewed (code-review + security-review panel: no bypass found); findings dispositioned

Spec: ACE-037

### DIFF
--- a/deploy/agami.env.example
+++ b/deploy/agami.env.example
@@ -56,3 +56,9 @@ AGAMI_ADMIN_PASSWORD=
 # AGAMI_REFRESH_TOKEN_MODE=overwrite # refresh-token storage: overwrite (default: one row/session, no
 #                                    # dead-token heap; DROPS stolen-token reuse detection) OR rotate
 #                                    # (keeps OAuth 2.1 reuse detection + prunes expired revoked rows)
+
+# Optional: SQL scopability posture. `enforce` (the default) REFUSES any query the executor can't
+# fully scope to your declared tables — unparseable SQL, or a table-function / VALUES / UNNEST /
+# LATERAL source. This is the safe default. `warn` logs and allows instead; use it only as a
+# temporary staged-rollout escape hatch, never as the shipped setting.
+# AGAMI_SQL_UNSCOPABLE_POSTURE=enforce
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -920,6 +920,14 @@ def _refusal_from_verdict(kind: str, verdict: Verdict) -> Refusal:
     return Refusal(kind=kind, reason=verdict.detail, remediation=verdict.remediation)
 
 
+def _unscopable_posture() -> str:
+    """The unscopable-SQL rollout posture: ``enforce`` (default — fail-closed) or ``warn`` (a
+    staged-rollout escape hatch that logs and allows). ``warn`` is never the shipped default; safety
+    fails closed. This is an operational rollout knob, NOT the deployment tier — safety has no tier
+    variance and enforces in every tier."""
+    return os.environ.get("AGAMI_SQL_UNSCOPABLE_POSTURE", "enforce").strip().lower()
+
+
 def _model_safety(sql: str, profile: str, area: str | None) -> tuple[str, Refusal | None]:
     """Semantic-model safety pass before execution: fan-trap / chasm-trap pre-flight
     + default_filters auto-application, over a model resolved from the DB (hosted) or disk (local).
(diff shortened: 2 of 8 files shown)


ashwin-agami (commented):
Why even have a flag to warn? Why not just make it strict? Where do we have the use case for warning?

sandeep-agami:
Closing: superseded, and its base is doubly stale.

Its base is `ACE-035-shared-guardrail-contract` — but ACE-035 shipped from a different branch entirely (`ACE-035-guardrail-contract-rewrite`, #170, `9d3fd8d`), because the contract was **reduced** rather than ported. `Verdict{class, severity, certainty}` and `policy(verdict, tier) → Action` are gone; what shipped is two dataclasses, `Refusal` and `Envelope`. This branch is built on the version that was deleted.

ACE-037 itself is still live and still needed — fail-closed on unparseable or unscopable SQL is Wave 3, and it now has a second reason to exist: ACE-039's residual is that on the vendored plugin layout `semantic_model.runtime` is absent by construction, so catalog relations have no gate there at all. Failing closed when the guard cannot run is the general form of that fix.

It gets a fresh branch off `main` when Wave 3 starts. What the re-authored version inherits is the mechanism — the scopability walk and the comma-join handling from `5a31525`/`a1e5065` — not this diff: the refusal is now returned rather than raised, `GuardRefused` no longer exists, and `RULE_UNSCOPABLE` is deliberately left unpinned in `REASON_FOR_RULE` for the gate that produces it to fill (`test_ace035_guardrail_contract.py` asserts exactly that).

sandeep-agami:
Recording why this stays closed. No action needed; this PR is not being reopened.

**Spec:** ACE-037 (fail-closed on unscopable SQL)

Treat this as a reference implementation, not a pending review. The work is being rebuilt against `main` under ACE-037.

Nothing is lost. This PR's head (`a1e5065`) is contained in the immutable tag `governance-branch-reference` (`76ed6cb`). `agami-governance-branch` is being deleted, so read reference code from that tag, never from a branch name.

PR #120, opened 2026-07-12
Recon deny-list + DB-error sanitization for the SQL guard

sandeep-agami:
## Summary
Closes two information-leak channels that remain open even under read-only + object-scope:

1. **Recon / metadata deny-list** — a new `sql_guard.check_no_recon` gate refuses queries that fingerprint the server (`version()`, `current_user`, `inet_server_addr()`), enumerate or **dump** the system catalog (`information_schema`, `pg_catalog`, any `pg_*` relation — including `pg_stats` sampled VALUES and `pg_authid` password hashes), read catalog DDL (`pg_get_viewdef`/`pg_get_functiondef`), or probe object existence (`has_table_privilege`, `to_regclass`). The read-only role can't revoke catalog access on most engines, so this is the intended app-side backstop. A distinct `recon` refusal, wired at the single executor chokepoint (`execute_sql.py::main`) + a fail-fast tool-layer pre-check, so both transports inherit it.

2. **DB-error sanitization** — a raw driver error returned to the caller leaks schema, column names, or data values across the assistant boundary. The tool layer now classifies an operational failure into a value-free message (rich `kind` mirroring `db_error_classifier.md`), and the raw detail goes **only** to a new `guardrail_audit.error_detail` audit column, keyed by `audit_id`.

## Changes
- `sql_guard.py`: `check_no_recon` + data-driven recon sets, regex over the same neutralized SQL as `check_read_only` (no second parser). `pg_` is a reserved relation/function prefix (mirrors the Redshift `stl_/…` prefixes); qualified columns (`t.pg_class`) pass via a `(?<!\.)` lookbehind. Fails **closed** when the SQL can't be neutralized.
- `execute_sql.py` / `tools.py`: wire the recon gate; `_classify_db_error` + `_refusal_from_stderr` returns `(Refusal, error_detail)`; the raw stderr is threaded to the audit row only.
- `guardrail.py`: add `network`/`column_not_found`/`table_not_found` to `REFUSAL_KINDS` (+ the drift test now enumerates every emitted kind).
- `model_store.py` / `store.py` / migration `013`: the `error_detail` column + a graceful INSERT degrade (survive a pre-013 schema instead of dropping the audit row) + `Store.rollback`.

## Verification
- Recon corpus enumerates **every** deny-list member (self-updating) + the `pg_` leak closures + neutralization bypasses (comment/case/quoted-identifier/set-op/subquery); a **false-positive** corpus pins the legit look-alikes (bare `version`/`user`/`schema` columns, `t.pg_class`, tokens in literals/comments, window fns/CTEs) with the accepted residuals pinned explicitly.
- Sanitization corpus asserts the raw token is absent from the **entire** envelope JSON across every kind + dialect + the high-leak host/user cases, and the audit both-ends (raw in `error_detail`, absent from the response).
- Full test gate green (**1655 passed**); driven end-to-end through the executor CLI (`pg_stats`/`pg_authid`/`pg_get_viewdef`/`has_table_privilege`/`inet_server_addr()` → `recon`; a qualified `o.pg_class` column returns data).

## Review
Ran the Agami review panel (correctness · silent-failure · test-quality · High-lane **security**). Error-sanitization sign-off satisfied; the panel found real primary-engine recon false-negatives (the `pg_` catalog relations + introspection functions) and a pre-013 audit-row-drop — all fixed here, with the corpus expanded to prove them closed.

## Checklist
- [x] `Spec: ACE-039`
- [x] Both surfaces (stdio + HTTP) via the single `execute_sql`/`tool_execute_sql` chokepoint
- [x] Vendored `plugins/agami/lib` in sync (`dev.py sync-lib`; drift-checked)
- [x] Full gate green; no raw DB text on any Envelope path (security-verified)
- [ ] Human PR approval

Stacked on #115; retarget to `main` once #111 / #112 / #115 merge.

Spec: ACE-039

### DIFF
--- a/migrations/core/013_guardrail_audit_error_detail.sql
+++ b/migrations/core/013_guardrail_audit_error_detail.sql
@@ -1,0 +1,11 @@
+-- Store the RAW driver/DB error text server-side, keyed by audit_id. The caller only ever receives a
+-- generic, value-free classified message (raw schema / column / value names must never cross the
+-- assistant boundary), so the operator needs somewhere to see the real error when debugging a failed
+-- query — that is this column. Populated ONLY on an operational failure (the executor's stderr);
+-- NULL for guardrail refusals (their reasons are already clean). Portable (SQLite + Postgres).
+--
+-- SENSITIVITY: a driver error can embed data VALUES (e.g. a Postgres unique-violation
+-- `DETAIL: Key (email)=(...)`), so this column inherits the datasource's sensitivity classification.
+-- It is strictly better than the prior behavior (which returned that text to the caller), but for a
+-- PHI/PII deployment it must fall under the audit table's access-control + retention/redaction policy.
+ALTER TABLE guardrail_audit ADD COLUMN error_detail TEXT;
--- a/packages/agami-core/src/contracts.py
+++ b/packages/agami-core/src/contracts.py
@@ -217,3 +217,4 @@ class GuardrailAuditRecord(_Contract):
     execution_ms: int | None = None
     correlation_id: str | None = None
     source: str | None = None
+    error_detail: str | None = None  # RAW driver error — server-side audit only, NEVER in the Envelope
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -1347,19 +1347,25 @@ def execute_guarded(
 ) -> ExecResult:
     """The un-bypassable guarded envelope — the single execution chokepoint (REQ-002/REQ-014).
 
-    In fixed order: read-only / dangerous-SQL guard (the hard security gate — NOT bypassable via
-    ``no_safety``, which skips only the semantic-model pass, never write/RCE/DoS protection) ->
-    semantic-model safety pass (fan/chasm pre-flight + scope + PII + ``default_filters`` rewrite) ->
-    resolve the datasource -> ``executor.execute(vetted_sql, …)``. The executor only ever receives
-    SQL both guards have passed. Raises ``GuardRefused`` carrying the typed ``Refusal`` on a refusal,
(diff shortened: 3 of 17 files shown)


sandeep-agami:
Superseded by #176, which ports this to `main` as a fresh branch.

This PR cannot merge to `main`: its base is `ACE-038-resource-limits-timeout`, a retired branch, so `MERGEABLE` here means mergeable into that. Merging it would drag 8 commits, including `3950f08`/`886428a` — the **old** ACE-035 `Verdict` / `policy(verdict, tier)` contract that the rewrite deleted and #170 replaced with `Refusal` / `Envelope`.

The port is a port of the *mechanism*, not of this diff. What changed on the way (all recorded in ACE-039's `## Decisions`):

- **Functions only.** The schema, relation and relation-prefix sets are gone; catalog relations are `check_table_scope`'s job. That also removes the over-refusal this branch caused in real use.
- **Register-type casts added**, both spellings and schema-qualified.
- **The refusal is returned, not raised** — `GuardRefused` no longer exists.
- **`reason` + `rule`**, not `kind`; `REASON_FOR_RULE[RULE_RECON] = "unsafe"`.
- **Database errors go to `Envelope.failure` with no remediation**, never to a refusal.
- **Cancellation signatures ceded to ACE-038**, kept only to route an unattributable cancellation to `other`.
- **An unreadable statement refuses `undetermined`/`unparseable`**, not `recon`.
- **`read_only` where this said `permission`.**
- **Two false positives fixed**, including the quoted-identifier one, which needed `_neutralize` to carry quote provenance — and had to be scoped to the niladic matcher only, or it would have reopened the `c42f96c` file-read bypass.
- Migration `013` renumbered to `016`; `main` had consumed the number.

Closing rather than merging, per the governance remediation plan §4.

PR #121, opened 2026-07-13
ACE-040: safety regression corpus — the F9 done-bar

sandeep-agami:
## Summary
`Spec: ACE-040` — the F9 done-bar. One adversarial **safety regression corpus** that drives every attack class through the REAL `execute_sql` tool and asserts its expected `Envelope`, on **both surfaces** (stdio subprocess + in-process HTTP) and **both model paths** (file-served: SQLite + disk YAML; DB-served: Postgres + `model_store`), plus a **read-only role floor** that proves the DB rejects a write with the app gate bypassed. This is the regression bar a future change must not drop below. Test-only + CI — no `src` changes; it locks the controls built in ACE-035…039.

Stacked on **#120 (ACE-039)** — base is `ACE-039-recon-error-hardening`; retarget to `main` once the stack lands.

## Changes
- **`tests/safety/corpus.py`** — the single canonical source: the demo `SCHEMA` (the harness derives BOTH the semantic model and the physical datasource from it) + `CASES` mapping each class to its outcome. Classes: integrity→`permission`, object-scope→`table_out_of_scope`/`select_star`/`column_out_of_scope` (incl. hidden in `UNION`/`EXCEPT` arms), fail-closed→`unscopable_sql`, recon→`recon`, availability→bounded (row-cap truncate+flag), governed→`ok` (no false refusals).
- **`tests/e2e/harness.py`** — ONE copy of the stdio + HTTP drivers + the model/datasource builders (`test_safety_envelope.py` now imports them — drivers single-sourced, no duplication).
- **`tests/e2e/conftest.py`** — `surface` (stdio|http) + `file_safety_env` + the Postgres fixtures (`pg_admin`, `create_ro_role`, `db_safety_env`, `pg_ro_conn`).
- **`tests/e2e/test_safety_corpus.py`** — the corpus over `{surface} × {file, DB}` paths.
- **`tests/e2e/test_role_floor_pg.py`** — writes on a RAW `agami_ro` connection (app gate bypassed) → Postgres raises `InsufficientPrivilege`; a `SELECT` succeeds.
- **`.github/workflows/ci.yml`** — a new `integration-pg` job (a `postgres:16` service — the repo's first CI service) runs the DB-dependent tests; an `AGAMI_IT_PG_REQUIRED` sentinel makes an unavailable DB **fail** (not skip) so an all-skip can't pass green.

## Verification
- **File path (local): green** — `uv run dev.py check` passes; 58 corpus cases (29 × both surfaces) + the envelope tests. Full suite 1752 passed / 62 skipped.
- **Postgres path (local): skips** — no Docker on this machine, so the DB-served corpus + role-floor **skip cleanly locally**; the `integration-pg` CI job is their first real run. The fixtures were written against the compose fixture's known creds and read as if debugging; a focused security review traced the guard order per case, confirmed the role-floor genuinely bypasses the app gate, and confirmed the corpus is non-vacuous (a regressed control degrades each case to a failure, never a silent pass).

## Reviewer notes / follow-ups
1. **Make `integration-pg` a required check** (branch protection — outside this diff). The sentinel guarantees it can't pass green on an all-skip, but it must actually be required to gate merges.
2. **PG path is first-run-in-CI** (no local Docker here) — worth a close look at the first `integration-pg` run.
3. **High-lane → security sign-off required** per the spec.
4. Consolidation is **additive** (per sign-off): the corpus is the new end-to-end bar + single-sourced vectors; the existing ~15 unit gate-tests stay as the fine-grained layer (full retire is a follow-up).

## Checklist
- [x] `Spec: ACE-040` on every commit; diff stays in spec scope (declined a PII/`sensitive_columns` case — spec assigns PII to F10's corpus)
- [x] Both surfaces + both paths exercised and asserted identically
- [x] No weakened/deleted tests; the one refactor (shared drivers) keeps `test_safety_envelope.py` green
- [x] `uv run dev.py check` green (file path); review dispositioned (1 must-fix fixed)
- [ ] `integration-pg` green in CI (first run) + made a required check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,57 @@ jobs:
           --with-editable "packages/agami-core[model,server]"
           pytest tests/ -q --cov=plugins --cov=packages/agami-core/src --cov-report=term-missing
 
+  integration-pg:
+    # The F9 safety regression gate that needs a real database: the Postgres-SERVED corpus path
+    # (file/db parity) + the read-only role floor. These skip cleanly in `lint-and-test` (no DB);
+    # here a Postgres service is present so they RUN. Make this a required check for the F9 done-bar.
+    name: integration (Postgres — F9 safety corpus + role floor)
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_DB: shop
+          POSTGRES_USER: agami_test
+          # Ephemeral localhost CI service: trust auth (no password) — nothing to leak, and the tests
+          # assert role PRIVILEGES, not authentication. Avoids a throwaway password literal in the diff.
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432:5432
+        # Wait for the server to accept connections before the steps run.
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+    env:
+      # The env-gated PG tests read these (defaults target the local compose fixture); point them at
+      # the service. AGAMI_IT_PG_PASSWORD being set is what flips the tests from skip to run.
+      AGAMI_IT_PG_HOST: 127.0.0.1
+      AGAMI_IT_PG_PORT: "5432"
+      AGAMI_IT_PG_USER: agami_test
+      # A non-secret sentinel: it only has to be non-empty to flip the tests from skip→run (the service
+      # uses trust auth, so the value is never verified). Not a credential.
+      AGAMI_IT_PG_PASSWORD: trust
+      AGAMI_IT_PG_DB: shop
+      # In this job a DB is REQUIRED: an unavailable/unreachable Postgres FAILS the job instead of
+      # skipping, so an all-skip can't pass green and silently prove nothing (the F9 done-bar gate).
+      AGAMI_IT_PG_REQUIRED: "1"
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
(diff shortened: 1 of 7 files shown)


sandeep-agami:
Closing: this PR's base branch no longer leads anywhere.

Its base is `ACE-039-recon-error-hardening`, which was superseded by #176 (merged as `2481c5f`) and whose own PR #120 is closed. So `MERGEABLE` here means mergeable into a retired branch, and merging it would drag the stacked ancestry back — including the **old** ACE-035 `Verdict` / `policy(verdict, tier)` contract that #170 replaced with `Refusal` / `Envelope`.

ACE-040 is not abandoned; it is **unblocked as of today**. Its two remaining dependencies both shipped: ACE-036 (`41cfd7d`) and ACE-039 (`2481c5f`). It gets a fresh branch off `main`, and the corpus it lands is the done-bar on every remaining wave rather than a one-off.

What the re-authored version inherits from here: the test files themselves, and the sentinel defect #161 identified — enforcement per call-site, so renaming one test silently drops 102 of 108 vectors. Per the governance remediation plan §4, that must be fixed **in** the port, not after it.

What it will need to add, because `main` moved underneath this branch: recon vectors (ACE-039), `resource_limit` vectors (ACE-038), and the four new `FailureKind` members with their exit codes.

sandeep-agami:
Recording why this stays closed. No action needed; this PR is not being reopened.

**Spec:** ACE-040 (safety regression corpus)

Treat this as a reference implementation, not a pending review. The work is being rebuilt against `main` under ACE-040.

Nothing is lost. This PR's head (`c8cb377`) is contained in the immutable tag `governance-branch-reference` (`76ed6cb`). `agami-governance-branch` is being deleted, so read reference code from that tag, never from a branch name.

PR #123, opened 2026-07-13
fix(onboarding): star-free seed probe, sm approve-queue + at-stamp, pytz-aware duckdb resolver (ACE-062)

ashwin-agami:
**Spec: ACE-062**

## Summary
Three reproducible defects surfaced by an external tester onboarding an OMOP CDM (Common Data Model) healthcare dataset — a schema where every clinical table carries `TIMESTAMP WITH TIME ZONE` columns — on the connect → curate → seed path. This is the **code** half; the docs/packaging half is **ACE-063** (depends on this PR).

## Changes
- **A1 — star-free seed probe** (`semantic_model/curate.py`). `validate_seeds()` wrapped each seed as `SELECT * FROM (<sql>) WHERE 1=0`; the wrapper's own `SELECT *` tripped `runtime.check_no_select_star` (which walks *every* SELECT in the tree), so **every** seed was rejected as `select_star` regardless of its SQL. Switched the wrapper to `SELECT 1` — still parses + plans the inner query and returns zero rows, but projects no star.
- **A2 — `sm approve-queue` + `at` auto-stamp** (`semantic_model/cli.py`). New `sm approve-queue <root> --signer --role [--kind ...] [--dry-run]`: reads the pending review queue (Rule 1 metrics/filters + Rule 2 joins/entities), builds a self-stamped `approve` op per item, and applies via `curate.apply` — a first-class **no-browser** sign-off path. Also fixes a silent trap: `curate.apply` reads `signed_off_at` from `op.get("at")` (the engine deliberately can't call the clock), so an approve op missing `at` recorded `signed_off_at: null` and the validator rejected the whole batch. `cmd_curate` (and `approve-queue`) now stamp `at` at the CLI boundary, where the clock is available. `--signer`/`--role` still required, so who-signed-off is unchanged.
- **A3 — pytz-aware duckdb probe** (`plugins/agami/scripts/connect_resolve.py`). duckdb needs `pytz` to materialize `TIMESTAMP WITH TIME ZONE` into Python; the probe only checked `import duckdb`, so an interpreter without pytz scored as ready and then failed at query time on any timestamptz column. Folded pytz into the duckdb `has_driver` probe via `_DRIVER_EXTRA` so the skill's driver-missing branch installs it up front.

## Test plan
- `test_validate_seeds_probe_is_star_free` — a legit named-column seed passes; the probe handed to the runner is star-free.
- `test_approve_queue_signs_off_all_pending` — queue → `total: 0`, `curate-gate` `preseed_count` → 0, sign-off fields recorded; `test_approve_queue_kind_filter_and_dry_run` — `--kind` narrows, `--dry-run` mutates nothing; `test_curate_auto_stamps_missing_at` — an approve op with no `at` now applies cleanly.
- `test_duckdb_driver_requires_pytz` — duckdb-without-pytz scores `has_driver: false`; with pytz, `true` and chosen.

## Checklist
- [x] `uv run dev.py check` green — full suite **1503 passed**, ruff clean, gitleaks clean
- [x] `uv run dev.py cover` — **100%** patch coverage on changed lines
- [x] Regression test per fix; no tests weakened or deleted
- [x] Customer-safety: neutral placeholders only (`you@example.com`), no customer/personal identifiers in code, tests, or comments
- [x] `/agami-sdlc:review` run — 0 must-fix, 0 nits
- [ ] Human PR approval

## Out of scope
- Docs/packaging half (`pip install duckdb pytz`, render-flag fix, sample-6B auto-approve, headless-sign-off docs) → **ACE-063**.
- No change to the `SELECT *` ban or the preseed gate's strictness for real databases.

### DIFF
--- a/packages/agami-core/src/semantic_model/cli.py
+++ b/packages/agami-core/src/semantic_model/cli.py
@@ -27,6 +27,7 @@ import json
 import os
 import re
 import sys
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Optional
 
@@ -38,6 +39,23 @@ from . import validator as V
 def _print_json(obj) -> None:
     json.dump(obj, sys.stdout, indent=2, default=str)
     sys.stdout.write("\n")
+
+
+def _utc_now_iso() -> str:
+    """ISO-8601 UTC (`...Z`) — the sign-off timestamp format the validator expects."""
+    return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def _stamp_approve_ops(ops: list) -> list:
+    """Stamp `at` on any `approve` op that lacks one. curate.apply deliberately can't read
+    the clock itself (`op.get("at")` — "caller stamps"); an approve op without `at` records
+    `signed_off_at: null` and the validator then rejects the whole batch. The CLI process
+    *can* read the clock, so it stamps here — one place, for every CLI caller."""
+    now = _utc_now_iso()
+    for op in ops:
+        if isinstance(op, dict) and op.get("op") == "approve" and not op.get("at"):
+            op["at"] = now
+    return ops
 
 
 def cmd_validate(args) -> int:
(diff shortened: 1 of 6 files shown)


ashwin-agami:
Thanks @Copilot — both addressed in 4f06577:

1. **`approve-queue` signer (must-fix):** made `--signer`/`--role` **required** on the subcommand. You're right that `curate._set_trust` only stamps `signed_off_*` when a signer is present, so a signer-less approve recorded an incomplete trust block and the validator reverted the whole batch. argparse now rejects the call up front, with a regression test (`test_approve_queue_requires_a_signer`).
2. **Brittle test (nit):** the star-free probe test's ban simulation now uses a case/whitespace-insensitive regex `(?i)\bselect\s*\*` (mirroring the AST-based `check_no_select_star`) instead of an uppercase substring — `SELECT  *` / `select *` would now be caught too.

PR #125, opened 2026-07-13
docs(onboarding): pytz install, render flags, sample-6B auto-approve, headless sign-off (ACE-063)

ashwin-agami:
**Spec: ACE-063** · docs/packaging half of the OMOP-onboarding fixes (follows #123).

## Summary
Four documentation / skill-flow gaps from the same onboarding session that produced #123. All docs-only; each documented command was verified against the real code.

## Changes
- **B1 — pytz in the DuckDB install.** Both driver tables now say `duckdb pytz` (skill 0a.5b table + `connection-reference.md`), with a one-line why: the DuckDB Python driver needs `pytz` to materialize `TIMESTAMP WITH TIME ZONE`, else a timestamptz query fails at runtime. Pairs with the resolver probe shipped in #123 (`_DRIVER_EXTRA`).
- **B2 — render_model_explorer invocation.** The sample 6B step named the script with a single directory, inviting a non-existent `--root`. Replaced with the explicit correct form `--profile agami-example --artifacts-dir <artifacts_dir> --out …` and a pointer to the canonical `/agami-model` snippet.
- **B3 — sample 6B clears its own pre-seed gate.** The "silent build" left metrics/entities `unreviewed`, so `seed-examples` refused with `preseed_review_pending`. Added a carve-out step to auto-approve the queue as `--signer system --role system` (via the new `sm approve-queue`) right before seeding. Real databases keep the human sign-off gate.
- **B4 — headless sign-off + PII advisory.** New "Sign-off without a browser" section in `/agami-model` documents `sm approve-queue` (required `--signer`/`--role`, `--kind`, `--dry-run`) as the no-browser path; both `/agami-model` and the connect gate now state explicitly that PII / `should_open_explorer: true` is **advisory** — it does not block seed generation; only unreviewed pre-seed items do.

## Test plan
Docs-only — no code paths change. Verified: `render_model_explorer.py` flags (`--profile`/`--artifacts-dir`/`--out`, no `--root`), `sm approve-queue` surface (required signer/role, `--kind`, `--dry-run`), and `_DRIVER_EXTRA` duckdb→pytz all match the shipped code (#123).

## Checklist
- [x] `uv run dev.py check` green — full suite passes, ruff clean, gitleaks clean
- [x] Every documented command/flag verified against the actual code (review agent)
- [x] Customer-safety: only `you@example.com` placeholder; no customer/personal identifiers, no real hosts, no `omop`
- [x] `/agami-sdlc:review` run — 0 must-fix
- [ ] Human PR approval

## Out of scope
- All code behavior (shipped in #123). No change to the `SELECT *` ban or the preseed gate's strictness for real databases.

### DIFF
--- a/plugins/agami/shared/connection-reference.md
+++ b/plugins/agami/shared/connection-reference.md
@@ -224,9 +224,11 @@ The "Python Driver Fallback" section further down shows the inline `python3 -c '
 | MySQL / MariaDB | 3306 | `mysql` | `pymysql` (`pip install pymysql`) | optional |
 | **Snowflake** | 443 (HTTPS) | `snowsql` | `snowflake-connector-python` (`pip install snowflake-connector-python`) | TLS always (managed by client) |
 | SQLite | N/A (file) | `sqlite3` | built-in `sqlite3` | n/a |
-| DuckDB | N/A (file) | `duckdb` | built-in or `pip install duckdb` | n/a |
+| DuckDB | N/A (file) | `duckdb` | built-in or `pip install duckdb pytz` | n/a |
 
 Supported end-to-end: Postgres + Redshift + MySQL + Snowflake + SQLite + **BigQuery**. SQLite also works via DuckDB. Other databases (SQL Server, Oracle, Databricks, ClickHouse) are deferred.
+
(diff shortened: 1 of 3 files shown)


ashwin-agami:
Thanks @Copilot — fixed in the latest commit: `sm review-queue` nests the count under `counts.total`, so the doc now says `counts.total: 0` (matching `curate.review_queue`'s `{"counts": {"total": …}}` shape).

PR #128, opened 2026-07-14
ACE-041: PII masking (mask-else-refuse)

sandeep-agami:
## Summary
Turns the refuse-only sensitive-projection guard into **mask-else-refuse** (F10 data-protection): a raw projection of a model-declared `sensitive` column is redacted to `***` instead of blocking the whole query; a projection that can't be safely traced to a single output column is **refused** (fail-closed). Enforced once at the shared executor chokepoint, so it holds on every engine and on both surfaces.

`Spec: ACE-041`

> **Stacked PR** — based on `ACE-035-shared-guardrail-contract` (#111), **not `main`**. Merge after #111.

## Changes
- `runtime.check_sensitive_projection` → per-projection `SensitiveProjection{column, maskable, output_index}` (bare / `t.col` / aliased → maskable at its output index; expression / function / scalar-subquery / `*` → must-refuse; every set-op arm walked).
- `guardrail.policy()` `data_protection` branch: `certainty="provable" → "mask"`, else `"reject"` (fail-closed, enforces in every tier).
- `execute_sql._model_safety` builds a `data_protection` `Verdict` (`provable` iff **every** offending projection is maskable) → `policy` → a `MaskPlan` or a refusal; `execute_guarded` applies redaction in a new **shared post-`executor.execute()` step** (covers any injected `ports.Executor`; composes with the already-merged row cap).
- `applied[{mask}]` recorded on **both** surfaces (in-process 4-tuple → `_finalize_execution`; subprocess `{"masked":[...]}` stderr line). Cross-position set-op notes report every masked output position.
- **Case-fold fix:** sensitive-name matching is now case-insensitive — closes a `SELECT "SSN"` / `SELECT SSN` bypass, consistent with the table/column-scope gates.
- **Unparseable-with-PII** fails closed (refuse) on this branch.

## Test plan
- Adversarial SQL matrix (maskable / must-refuse / allowed-unchanged / edge), each on both surfaces where applicable, plus a **real subprocess-fork e2e**, an **injected `Executor`**, numeric / NULL / empty results, and row-cap composition.
- `dev.py check` green: ruff (check + format), **1608 passed / 1 skipped / 2 xfailed**, gitleaks, vendored-lib drift exit 0.
- **No F9 safety test weakened** (safety suites 333 passed, unchanged); adversarial security review found no leak.

## Sign-off items (why this is a draft)
- [ ] **Security sign-off (High lane)** — this is the boundary between the LLM and the customer's database.
- [x] Redaction token confirmed `***`.
- [ ] **Known limitation — tracked as ACE-062:** a sensitive column renamed through a derived-table / CTE body (`WITH t AS (SELECT ssn AS s FROM customers) SELECT s FROM t`) escapes both mask and refuse. **Pre-existing** in the name-based sensitive-projection gate (present on the base branch), **not introduced here**. Documented in-code + pinned by `@pytest.mark.xfail` tripwire tests; follow-up spec **ACE-062** filed.

Operator note: `--no-safety` skips masking along with all model guards (operator-only; not reachable from the MCP tool).

### DIFF
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -1248,6 +1248,35 @@ def execute_guarded(
     return result
 
 
+def _mask_note(
+    declared: tuple[str, ...], indices: tuple[int, ...], out_columns: list[str]
+) -> tuple[str, ...]:
+    """The ``applied[{mask}]`` receipt for a redaction — accurate about EVERY masked output position.
+
+    ``declared`` are the model's sensitive "table.column" refs the gate named (``MaskPlan.columns``);
+    ``indices`` are the redacted 0-based output positions; ``out_columns`` are the executor's actual
+    output labels (``result.columns``), which for a set operation come from its FIRST arm. A single
+    sensitive projection redacts exactly one position whose own label IS that sensitive column, so the
+    note is just its declared ref (the spec's success criterion 1 — e.g. ``("customers.ssn",)``).
+
+    A CROSS-POSITION set operation (e.g. ``SELECT ssn, name … UNION ALL SELECT name, ssn …``) redacts
+    a SECOND output position whose first-arm label is a non-sensitive name (``name``) — necessary,
+    because that merged output column carries a raw SSN in the second arm's rows. Naming that extra
+    position by its output label keeps the receipt honest: it reflects every redacted position, not
+    only the declared columns. A position is "already named by a declared sensitive column" iff its
+    output label case-folds to the bare column name of one of the ``declared`` refs; otherwise its
+    label is appended. This never DROPS a masked position (fail-safe for a receipt) and never renames
+    the single-column case (that position's label equals its declared column's bare name)."""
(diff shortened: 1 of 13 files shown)


sandeep-agami:
Closing: **ACE-041 is abandoned.**

The governance principles agreed 2026-07-30 include *"Agami holds no access policy of its own — it reads as the connection reads"* and *"the result comes back exactly as the database produced it."* Masking is a policy we would be holding, applied by modifying a result we had already read, so both halves of mask-else-refuse are gone.

What replaces it is not a smaller version of it. A column that must not be readable is not declared in the semantic model, and any statement reaching it is refused as out-of-scope — no new machinery. Where per-user redaction is genuinely needed, that is the warehouse's own masking policy applied to the connecting identity, which is what hosted and enterprise buy by carrying the end user's identity on the connection rather than one shared role.

The spec carries the full reasoning in its Decisions. ACE-062 and ACE-080, which built on this, are abandoned with it.

PR #156, opened 2026-07-28
ACE-072: cross-datasource bridge as a validated model type

ashwin-agami:
Spec: ACE-072

## Summary
First slice of **F16 — cross-datasource metrics**. Promotes the cross-datasource "bridge" (the shared-key link that says a customer in one datasource is the same entity in another, e.g. `accounts.account_key` = `customers.account_key`) from a skill-side file into a **first-class, validated semantic-model type** on the deployment-level `OrgRecord`. This is the foundation the cross-datasource *metric* (ACE-073) will reconcile through. Previously the bridge lived only in `<artifacts_dir>/local/cross_profile_relationships.yaml` — unvalidated, and lost on a machine re-onboard (the confirmed Northpeak v2 regression).

Scope is the **foundation only**. The idempotent "link your databases" offer + the governed `sm bridges` CLI are deliberately **out of scope → ACE-078**; the metric type/execution/browse are ACE-073/074/076.

## Changes
- **`models.py`** — new standalone `CrossDatasourceRelationship` (datasource+`schema.table`+column-list endpoints; `executable` restricted to `split`/`informational`, rejecting `same_engine`; optional cardinality; trust block; `endpoint_key` for dedup). Adds the `cross_datasource_relationships` bucket to `OrgRecord`.
- **`org_record.py`** — `load_org_record` now merges bridges from three sources (inline · sidecar `cross_datasource_relationships.yaml` · legacy `cross_profile_relationships.yaml` migration), deduped by endpoint tuple, **in-memory / read-only**, and **lenient** (a malformed optional file is skipped per-entry, never raised — preserving the runtime-path contract).
- **`validator.py`** — new deployment-level `validate_deployment()` pass: loads every attached datasource model and rejects a bridge that is `same_engine` or whose datasource/dataset/column endpoint doesn't resolve (`cross_datasource_executable_mismatch`, `cross_datasource_endpoint_unresolved`), fail-closed.
- **`cli.py`** — `sm validate` also runs the deployment pass when an `OrgRecord` is present (no new subcommand; a no-op on pre-F16 layouts).
- **Tests** — new `tests/test_cross_datasource_bridge.py` (model, validation reject/accept, idempotent + precedence-correct migration, malformed-file leniency, distinct unreadable-profile message, CLI wiring).

## Test plan
- `uv run dev.py check` green (1576 passed / 1 skipped; ruff + gitleaks clean). `uv run dev.py cover` = 100% patch coverage.
- Reviewed by the Agami review panel (correctness + security + structural, high lane): no stop-the-line issues; all findings addressed (leniency contract, endpoint-message accuracy, exact-schema preference, test rigor).

## Checklist
- [x] Spec-linked (`Spec: ACE-072`), scope-contained (no ACE-073/074/078 work)
- [x] Public-safe: synthetic `acme`/`demo`/`account_key` fixtures only — no customer or internal naming
- [x] New behavior covered by tests; CI-equivalent green locally
- [x] Fail-closed validation; no SQL-guard surface touched (model + validation only)


ashwin-agami:
Thanks @copilot — both findings were valid and are fixed in 2ade01a.

`_load_bridges` now extracts entries via an explicit shape check (`_bridge_entries`): a `relationships:` mapping, a bare YAML list, or `[]` — so a bare-list bridge file no longer calls `.get` on a list. The read/parse of the optional sidecar + legacy files was also wrapped in an earlier commit (bc87c50), so a corrupt/unreadable file degrades to no bridges rather than raising on the runtime-path `load_org_record`.

Both are locked by new tests: `test_bare_list_sidecar_loads` and `test_corrupt_bridge_yaml_file_does_not_raise`.

PR #157, opened 2026-07-28
Read every statement in the engine's own grammar, and refuse when we cannot

sandeep-agami:
## Summary

The model-scoping guards decide by inspecting a parsed statement, and they parsed every statement in sqlglot's **generic** grammar — which matches no real engine. On a backtick-quoting datasource (MySQL, BigQuery, Databricks) `` SELECT `ssn` FROM `customers` `` parsed to **no tables and no columns**, so table scope, column scope and the sensitive-projection gate each found nothing to object to and allowed it. The column came back verbatim. Measured on the base branch:

```
MySQL datasource, ssn declared sensitive
  undeclared backtick table refused?   False
  undeclared backtick column refused?  False
  sensitive ssn noticed?               False
  receipt tables_used:                 []
```

The read-only lexer is a separate code path and still blocked writes, so this is **disclosure, not integrity**.

Backticks were the first instance, not the class. SQL Server `TOP` + `[brackets]` parsed to a *confidently wrong* tree — zero tables plus a phantom column named `TOP`, so a gate refused for a reason unrelated to the query — and a MySQL `#` comment truncated the parse.

Compounding it, **`error_level` was inert**. sqlglot compares the level against `ErrorLevel` members, so the string these calls passed matched no branch and every collected parse error was silently discarded. Worse, the constructor reads `error_level or ErrorLevel.IMMEDIATE`, so passing the string *overrode a safe default with a value matching nothing*. Swapping `"ignore"` → `"raise"` would have changed nothing and would have passed review looking like hardening.

Every guard-path parse and every SQL regeneration now takes the datasource's dialect and raises on parse failure.

## Changes

**The fix**
- New `semantic_model/sql_dialect.py`: an explicit `StorageType` → sqlglot map, one entry per engine. Hand-written because lowercasing is *not* the mapping — `PostgreSQL` → `postgresql` and `SQLServer` → `sqlserver` are both rejected by sqlglot (it wants `postgres` and `tsql`), and those are the most common engine and the most divergent quoter. Resolution raises rather than falling back to generic.
- All 11 parse sites in `runtime.py` funnel through one helper passing `dialect=` + `ErrorLevel.RAISE`, catching both `ParseError` and `TokenError`. Five duplicated try/except blocks are deleted, so this is a net simplification.
- All **three** emission sites regenerate in the datasource's grammar — including `pre_flight_check`'s `arm.sql()`, which the spec's rewrite list missed. These emit the statement the executor runs; emitting generically hands it SQL the guards never validated.

**Fail closed**
- A datasource whose engine cannot be determined refuses, whatever the model declares and whichever gate would otherwise have run — deliberately **ahead of `AGAMI_SQL_UNSCOPABLE_POSTURE`**, which is a staged-rollout hatch for unscopable *queries*, not for datasources nobody can parse for.
- An **unreadable statement** refuses ahead of that switch too. This is not inherited: raising parse errors *widens* what escapes through `warn`, because the old truncated tree still named enough for table scope to catch some statements, while `tree is None` makes every downstream gate degrade to allow.
- A dialect-independent backstop: a statement that reads from something yet resolves to no named table is refused — the shape a mis-read produces, caught without waiting for the map to be complete.
- Two distinguishable remediations: the regenerable case names a query-level next move; the unmapped-engine case names the operator action and does **not** invite a retry.

**Three holes found during review, all closed here**
1. **Row scoping was fail-open.** When a table's `default_filters` could not be applied, `apply_default_filters` returned the statement unchanged with an empty `applied` list — and the caller keeps the rewrite only `if applied`, so "could not apply" was indistinguishable from "no filter needed". The query ran returning exactly the rows the filter exists to withhold. Now raises; both callers refuse as `model_unavailable`.
2. **The mirror of this defect.** On backtick engines sqlglot reads `"x"` as a *string literal*, so a double-quoted sensitive column resolved to zero columns — where the old generic parse had read it as an identifier and caught it. A regression this change would otherwise have introduced. MySQL under `ANSI_QUOTES` really does treat it as an identifier, so meaning depends on server configuration the guard cannot see. sqlglot does not preserve the quote character (`"x"` and `'x'` arrive identical), so the ambiguity is detected by re-reading under an ANSI-quoting grammar and comparing attributed columns. A genuine single-quoted literal is unaffected.
3. **Nothing reconciled the declared engine with the credentials.** The guard picks its grammar from the model; the executor picks its driver from `creds["type"]`. A mis-declared model had the guard vet — and, where a rewrite fires, *regenerate* — a statement in a grammar the database does not speak. Compared at the chokepoint and refused.

**Mechanical guards**
- An AST scan fails the build on any parse that omits its grammar or passes a *string* error level — the failure is invisible at review, and `error_level="raise"` reads like a fix. Covers sqlglot's whole parse family and our own helpers. Verified by injecting a bad parse.
- Tier 1 asserts an explicit dialect for every `StorageType` via `get_args`, as **set equality**, so a twelfth engine fails CI rather than silently parsing generically.
- The corpus gains backtick and bracket forms, so the guarantee runs over both surfaces rather than only in unit tests.

## Evidence

- **Refusal delta: 0 of 28** valid statements (14 corpus + 14 seeded examples).
- **sqlglot floor unchanged at `>=20`, deliberately.** The intended bump was to guard against `RAISE` strictness drift. Measured on **20.0.0, 23.0.0, 25.0.0, 26.0.0, 30.10.0**: all eleven dialect names resolve, all three attack statements parse correctly, and `RAISE` raises identically on every one. No drift to defend against, so no dependency bump.
- Source diff **+279/−56**; the rest is tests.

## Checklist

- [x] `ruff check` / `ruff format --check` / gitleaks pass
- [x] `dev.py sync-lib` — vendored lib in sync
- [x] No test weakened or deleted; three fixtures corrected to declare the engine they actually run against
- [x] Synthetic examples only; no spec ids in code, comments or commit messages
- [x] **`dev.py check` green** — 2272 passed, 0 failed
- [x] **Postgres integration green** — 180 passed (both surfaces × both model paths + read-only role floor), verified against a real `postgres:16` container

## Also in this PR: the MCP SDK ceiling

`mcp 2.0.0` was published **today at 13:45:28Z**. It removes `Server.list_tools`, which `mcp_http` builds its tool surface with, so every HTTP request fails at import. The declared floor had no ceiling, so a fresh resolve picked it up: the last CI run before that release passed (13:35Z), the next one failed (22:07Z). It would have broken every PR in this repo.

The quiet part is what it did to the safety corpus — half of every case runs over HTTP, so those cases erroring meant the guarantee they assert **was not being checked at all**. Coverage that looks present and tests nothing. Capped at `mcp>=1.2,<2`; raise it deliberately once `mcp_http` is ported.

With the cap: full suite 109 failures → **0**, Postgres integration → **180 passed**. That is also what makes the "on both surfaces" criterion genuinely verified here rather than asserted.

## Known limitations — stated rather than implied

- **Execution parity is unproven on eight of eleven engines.** Tiers 1/2/3a prove the *verdict* everywhere with no database, but only PostgreSQL, SQLite and DuckDB prove the executor runs what the guard vetted — all `"`-quoting, i.e. exactly the family where this defect cannot manifest. The per-engine execution matrix is separate work.
- **The refusal-delta corpus is in-repo test SQL, not LLM-emitted SQL.** It measures the shapes we already knew to write down, not what the model generates against a real semantic model — which is the population that actually matters for over-refusal.
- `_semi_additive_columns` skips an unparseable metric binding rather than refusing: it enriches an advisory check and withholds nothing, unlike the row-scoping path.
- Semi-structured access (`SELECT payload:ssn`) stays invisible to the gates under both grammars — a separate gap in what the gates detect.

Spec: ACE-079

### DIFF
--- a/packages/agami-core/pyproject.toml
+++ b/packages/agami-core/pyproject.toml
@@ -37,7 +37,11 @@ model = [
 # The HTTP MCP transport (mcp_http): the official MCP SDK + an ASGI server. Kept out of the base
 # install so the local executor/skill stay stdlib-lean — only a hosted deployment installs it.
 server = [
-  "mcp>=1.2",
+  # Capped below 2.0: that release removed `Server.list_tools`, which `mcp_http` builds the tool
+  # surface with, so every HTTP request fails at import. With no ceiling here a resolve simply
+  # picks it up and the whole HTTP transport — and the safety corpus that runs over it — breaks
+  # without a code change. Raise the cap deliberately, once mcp_http is ported to the new API.
+  "mcp>=1.2,<2",
   "anyio>=4",           # async_offload.run_blocking offloads blocking work off the loop (ACE-048)
   "uvicorn>=0.30",
   "psycopg2-binary>=2.9",
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -1345,7 +1345,19 @@ def _model_safety(
         else:
             return sql, Refusal("sensitive_columns", sens.reason, sens.suggestion or ""), None
 
-    new_sql, applied = RT.apply_default_filters(sql, org, area=area, ctx=ctx)
+    try:
+        new_sql, applied = RT.apply_default_filters(sql, org, area=area, ctx=ctx)
+    except RT.RowScopingUnavailable as exc:
+        # The model declares a row filter for a table this query reads, and it could not be
+        # applied. Running anyway would return rows the row scoping exists to withhold, so
+        # this refuses. The fault is in the model, not the query, so the remediation is the
+        # operator's and the statement is not worth re-emitting.
+        return sql, Refusal(
+            "model_unavailable",
+            str(exc),
+            "Correct the table's default_filters in the model so the row filter can be "
+            "applied for this datasource's engine.",
+        ), None
     if applied:
         # default_filters only add WHERE conditions, never change the projection list, so the mask
         # plan's output indices stay valid against the rewritten SQL's result.
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -1229,6 +1229,23 @@ def _model_safety(
     # returns the same verdict as one that builds its own.
     ctx = RT.build_guard_context(sql, org)
 
+    # Ungovernable-engine gate — if the datasource does not determine which engine its SQL
+    # runs on, there is no grammar to read the statement in, so no gate below can reach a
+    # trustworthy verdict and the statement is refused rather than run blind. This is
+    # deliberately ahead of, and not subject to, the unscopable posture switch: the posture
+    # is a staged-rollout hatch for queries that cannot be scoped, whereas a datasource that
+    # cannot be parsed at all is a configuration fault the operator must fix. Refusing here
+    # also keeps the guarantee independent of which gate would otherwise have run.
+    if ctx is not None and ctx.dialect is None and ctx.parse_error:
+        return sql, Refusal(
+            "unscopable_sql",
+            ctx.parse_error,
+            # No rewrite of the query helps, so this must not invite a retry — it names the
+            # one change that makes the datasource governable.
+            "Declare the datasource's engine (storage_connections[].storage_type) so its SQL "
+            "can be parsed for the right engine, then retry.",
+        ), None
+
     # Scopability gate — refuse a query that can't be fully scoped (unparseable, or a non-`Table`
     # FROM/JOIN source the scope walk can't reject) rather than run it blind. Runs
     # BEFORE the object-scope gates so an unscopable query fails closed instead of reaching their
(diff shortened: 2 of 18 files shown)


sandeep-agami:
Both addressed in 9b25955.

**Scan matching `.parse()` too broadly** — correct, and the code contradicted its own comment. `parse` now requires sqlglot as the receiver; the three distinctive names (`parse_one`, `transpile`, `maybe_parse`) still match however they are called. Added a table of look-alikes (`argparser.parse`, `self.parse`, a local `parse(text)`) pinning the distinction — the failure mode here is a scanner that cries wolf and gets exempted into uselessness.

**Fixture referencing a connection it never declares** — also correct, and the reasoning is the part worth keeping: it is harmless *today* precisely because nothing resolves the table-to-connection link, which is exactly why it would have gone unnoticed if something started to. The first connection is now named `c` to match the tables.

sandeep-agami:
Re-requested a Copilot pass on `0e25c10`, which adds one clarification on top of the two fixes.

Checking the fixture's table-to-connection link across every shape it builds:

```
('PostgreSQL',)          declared=['c']        referenced=['c']  dangling=None
('PostgreSQL', 'MySQL')  declared=['c','c1']   referenced=['c']  dangling=None
()                       declared=[]           referenced=['c']  dangling=['c']
```

The last row is intentional and now says so in the code: passing no storage types models a datasource that declares no engine, which is the state these tests exist to refuse. `Table.storage_connection` is required, so the table necessarily names a connection that is not declared — that dangling reference is the condition under test, not leftover inconsistency.

PR #160, opened 2026-07-29
ACE-083: classify traps by what a fan-out can actually inflate, and give the scope walk a scope

### LINKED ISSUE #151: execute_sql fan-trap guard false-refuses duplication-insensitive aggregates over a one-side column: COUNT(DISTINCT pk), MIN/MAX, and mixed-grain expressions (preflight_refused / risk: fan_trap)
## Summary

The `execute_sql` fan-trap pre-flight guard (`preflight_refused`, `risk: fan_trap`) refuses (or silently auto-rewrites) a query the moment **any column from the one-side table appears inside an aggregate function**, regardless of *which* aggregate function it is or whether that column is actually the measure. A fan-out (one-to-many join) only inflates aggregates that are **sensitive to row duplication** — `SUM`, `AVG`, plain `COUNT`. It cannot change a **duplication-insensitive** aggregate: `COUNT(DISTINCT <one-side pk>)`, `MIN(...)`, `MAX(...)` are invariant under repeated rows. The guard does not make that distinction, so it wrongly refuses a set of extremely common analytical queries.

To be clear about scope: the guard already attributes an aggregate to its source table (`_aggregate_source_tables`) and correctly **allows** the common star-schema pattern of summing a *many-side* fact across a join (`SELECT c.region, SUM(o.amount) ... GROUP BY c.region` passes today, and there is a test for it). This issue is the **residual precision gap** on top of that: the classification keys off *"a one-side column appears inside an aggregate"* rather than *"a duplication-sensitive rollup of a one-side measure."*

The same class of over-refusal as #148 and #150: the guard acts before the SQL has been analyzed precisely enough to justify the action.

## Environment

- Component: `packages/agami-core/src/semantic_model/runtime.py` (`pre_flight_check` → `_preflight_select`), surfaced by `packages/agami-core/src/execute_sql.py`.
- Reproduced directly against `pre_flight_check` on current `main` (see below — every classification is the actual output).

## Reproduction (generic tables)

Model: `customers` (one) → `orders` (many) on `orders.customer_id = customers.id`.

### False positives — these are refused today but are provably exact

```sql
-- FP1: distinct count of the one-side PK across the fan-out.
-- COUNT(DISTINCT) de-duplicates, so the fan-out cannot inflate it.
SELECT c.region, COUNT(DISTINCT c.id) AS n
FROM customers c JOIN orders o ON o.customer_id = c.id
WHERE o.status = 'won'
GROUP BY c.region;
-- actual: preflight_refused, risk=fan_trap  (WRONG — should run)

-- FP2: MIN/MAX of a one-side column. Idempotent under row duplication;
-- repeating a customer's row N times does not change MIN/MAX.
SELECT c.region, MIN(c.signup_date)
FROM customers c JOIN orders o ON o.customer_id = c.id
GROUP BY c.region;
-- actual: preflight_refused, risk=fan_trap  (WRONG — should run)

-- FP3: mixed-column aggregate. The measure is per-order (many side);
-- the one-side column is only a scalar co-factor. Grain is the many side,
-- so there is no duplication of a one-side measure.
SELECT SUM(o.amount * c.fx_rate)
FROM customers c JOIN orders o ON o.customer_id = c.id;
-- actual: preflight_refused, risk=fan_trap  (WRONG — should run)
```

### Genuine fan trap — correctly flagged, and should stay flagged

```sql
-- A SUM of a ONE-side additive measure across the fan-out really does
-- multiply by the child count (each customer row repeats once per order).
SELECT SUM(c.employee_count)
FROM customers c JOIN orders o ON o.customer_id = c.id;
-- actual: risk=fan_trap, action=auto_rewrite  (CORRECT)
```

The inflation here is real (illustrative magnitudes): `SELECT SUM(employee_count) FROM customers` might be ~1.6M, while the same sum across the join is ~16x that, because each customer is repeated once per order. Fan traps are a genuine hazard — the intent of the guard is right; only its precision is wrong.

### Already correct on `main` (shown to scope this issue)

```sql
-- Summing the MANY-side fact across the join is exact and is allowed today.
SELECT c.region, SUM(o.amount)
FROM customers c JOIN orders o ON o.customer_id = c.id
GROUP BY c.region;
-- actual: action=allow  (CORRECT — this is NOT the bug)
```

## Observed vs expected

| Query | Observed | Expected |
|---|---|---|
| `COUNT(DISTINCT one_side.pk)` across join | `preflight_refused` fan_trap | allow (distinct de-dupes) |
| `MIN`/`MAX(one_side.col)` across join | `preflight_refused` fan_trap | allow (idempotent under duplication) |
| `SUM(many_side.measure * one_side.factor)` | `preflight_refused` fan_trap | allow (measure grain is the many side) |
| `SUM(one_side.additive_measure)` across join | fan_trap (rewrite/refuse) | fan_trap — keep |
| `SUM(many_side.fact)` grouped by one-side dim | allow | allow — keep |

## Root cause

`_preflight_select` FAN branch — `packages/agami-core/src/semantic_model/runtime.py:960-1006`:

```python
# FAN: an aggregate over a measure on the ONE side of a one-to-many in scope
for measure_table in agg_sources:
    others = table_set - {measure_table}
    fan_rels = _one_side_facing_many(rels, measure_table, others)
    if not fan_rels:
        continue
    ...
    # -> refuse (or auto_rewrite): "fan trap: aggregating '<one>' (one side)
    #    across a join to [<many>] (many side)."
```

`agg_sources` comes from `_aggregate_source_tables` — `runtime.py:1397-1406`:

```python
def _aggregate_source_tables(tree, scope):
    """Tables whose columns appear inside an aggregate function in the SELECT."""
    sources = set()
    for select_expr in tree.expressions:
        for agg in select_expr.find_all(exp.AggFunc):
            for col in agg.find_all(exp.Column):
                t = _resolve_col_table(col, scope)
                if t:
                    sources.add(t)
    return sources
```

Two precision gaps follow from this:

1. **No aggregate-function-class check.** A table is treated as a fan-trap "measure on the one side" if *any* of its columns appears inside *any* `exp.AggFunc`. The specific function is never inspected, so `COUNT(DISTINCT c.id)`, `MIN(c.x)`, `MAX(c.x)` — all invariant under row duplication — are treated identically to `SUM(c.x)`. The one-to-many cardinality is then confirmed by `_one_side_facing_many` (`runtime.py:861-870`) and the refusal is emitted at `execute_sql.py:966-972`.

2. **No measure-vs-co-factor distinction.** Because *any* column inside the aggregate contributes to `agg_sources`, a one-side column used only as a scalar factor of a many-side measure (`SUM(o.amount * c.fx_rate)`) pulls the one-side table into `agg_sources`, even though the aggregate's grain is the many side.

The decision is therefore made from *(one-side column appears in some aggregate) × (one-to-many cardinality)*, not from *(a duplication-sensitive rollup of a one-side measure) × (one-to-many cardinality)*.

## Impact

Refuses several of the most common analytical shapes, e.g.:
- "How many distinct customers ordered, by region" (`COUNT(DISTINCT one_side.pk)` grouped by a joined dimension).
- "Earliest/most-recent one-side attribute among the joined population" (`MIN`/`MAX` of a one-side column).
- Currency/unit conversions and any measure scaled by a one-side rate (`SUM(many_side.measure * one_side.factor)`).

Tightening the classifier here also removes a chunk of user-visible `preflight_refused` noise without weakening the guard against real fan traps.

## Suggested fix directions

Only treat a one-side table as a fan-trap measure source when the aggregate over it is **duplication-sensitive**:

- **Aggregate-function class**: exempt `COUNT(DISTINCT <one-side col>)`, `MIN(...)`, `MAX(...)` (and typically `bool_and`/`bool_or`, `every`) — these cannot change under a fan-out. Keep `SUM`, `AVG`, plain `COUNT`/`COUNT(*)`, `STDDEV`, etc. as duplication-sensitive.
- **Measure vs co-factor**: only count a one-side table as a measure source when a one-side column is the aggregated *value*. When the aggregate expression also references a many-side column (mixed-column expression whose grain is the many side), do not classify it as a one-side fan trap.
- **Pre-aggregated inputs**: an aggregate over a subquery/CTE that is already grouped to the one-side grain is not a fan trap (this case already passes today; keep it a documented exemption so a future refactor does not regress it).

## Suggested regression tests

Add alongside the existing `test_aggregating_many_side_is_allowed` in `tests/test_semantic_model_runtime.py`:

- `COUNT(DISTINCT one_side.pk)` across a one-to-many join → `allow`.
- `MIN(one_side.col)` / `MAX(one_side.col)`, grouped by a joined dimension → `allow`.
- `SUM(many_side.measure * one_side.factor)` → `allow`.
- Guardrail (must stay red): `SUM(one_side.additive_measure)` across the join → `fan_trap`.

## Related

- #148 — `default_filters` rewrite corrupts valid SQL (scope-blind / dialect-blind).
- #150 — column-scope guard false-refuses valid date-math SQL.

Same underlying theme: analyze the SQL precisely before a guard refuses or rewrites it. The fan-trap hazard is real (the ~16x inflation above is genuine); the defect here is precision, not intent.

sandeep-agami:
**Spec: ACE-083**

The fan-trap detector decides from *(a one-side column appears in some aggregate) × (one-to-many cardinality)* rather than from *(a duplication-sensitive rollup of a one-side measure at one-side grain) × (one-to-many cardinality)*. Three fixes restore the missing terms. Closes #151.

## What was wrong

Measured through `pre_flight_check` against the test fixture model:

| query | before | why it is wrong |
|---|---|---|
| `COUNT(DISTINCT orders.id) FROM orders JOIN order_items …` | `fan_trap` | `DISTINCT` de-duplicates; a fan-out cannot inflate it |
| `MAX(orders.total_amount)` over the same shape | `fan_trap` | idempotent under row duplication |
| `SUM(order_items.quantity * orders.total_amount)` | `fan_trap` | the aggregated value is at many-side grain; the one-side column is a scalar co-factor |
| pre-aggregated CTE, then join | `fan_trap` | the CTE is grouped to the join key, so the join is 1:1 — **and this is the shape the refusal message itself recommends** |
| `WITH o AS (SELECT * FROM orders) SELECT SUM(o.total_amount) FROM o JOIN order_items …` | `allow` | genuinely inflated, and it passed |

The last two share one root cause: `_tables_in_scope` was a flat `find_all(exp.Table)` over the whole tree with no concept of scope, so CTE-body tables leaked into the outer query (false refusals) while CTE aliases never resolved to their underlying table (false allows).

## Changes

- **Fan-immune aggregates** — `MIN`, `MAX`, `DISTINCT`-qualified aggregates and boolean folds no longer feed the fan rule. Matched on the sqlglot node type and the `distinct` arg, not a name string, so it survives a dialect spelling it differently. `SUM`, `AVG`, `COUNT(*)`, `COUNT(col)`, `STRING_AGG` and `ARRAY_AGG` still trip it — one test each so the exemption cannot silently widen. `AVG` is the one worth naming: a fan duplicates rows *unevenly*, so averages do shift.
- **Measure vs co-factor** — an aggregate is attributed to the one side only when a one-side column is the aggregated value.
- **Scope** — `_tables_in_scope` is outer-scope only; a CTE alias resolves to its underlying table only when the CTE preserves grain (no `GROUP BY` / `DISTINCT` / aggregate), transitively; a grain-changing CTE keeps its own grain instead of inheriting the source table's cardinality; anything unclassifiable yields `certainty="uncertain"` rather than a silent `allow`.

This changes **what counts as a trap**, never what happens once one is found. The `auto_rewrite` branch is untouched.

## Verification

`dev.py check`: **2304 passed, 95 skipped, 2 xfailed**; ruff + gitleaks clean; no vendored-lib drift.

Collected-test delta is exactly the new cases: **2377 → 2401**.

Red-on-base, measured rather than asserted — base source with the new tests: **16 failed, 8 passed**. With the fix: **24 passed**. The 8 passing on base are the guardrail assertions (the fan-sensitive aggregates that must *still* refuse), which are supposed to be green on base.

### DIFF
--- a/packages/agami-core/src/semantic_model/runtime.py
+++ b/packages/agami-core/src/semantic_model/runtime.py
@@ -492,6 +492,16 @@ def resolve_entity_instance(
 
 @dataclass
 class PreFlightResult:
+    """The fan/chasm verdict, plus how well the analysis could see the query.
+
+    `certainty` uses the shared guardrail vocabulary: `"provable"` means every aggregate in the
+    statement was traced to a table the model declares, so `action` is a statement about the
+    QUERY; `"uncertain"` means at least one aggregate's grain could not be established (it reads
+    a CTE whose body the scope walk cannot classify), so `action` is only a statement about what
+    the analysis SAW. An `allow` carrying `"uncertain"` is "I could not tell", not "this is fine",
+    and the policy layer — not this gate — decides what that is worth.
+    """
+
     risk: Optional[str]  # "fan_trap" | "chasm_trap" | None
     action: str  # "auto_rewrite" | "refuse" | "allow"
     original_sql: str
@@ -499,6 +509,7 @@ class PreFlightResult:
     reason: str = ""
     suggestion: Optional[str] = None
     triggering_joins: list[str] = field(default_factory=list)
+    certainty: str = "provable"  # "provable" | "uncertain"
 
     def as_dict(self) -> dict[str, Any]:
         return {
@@ -509,6 +520,7 @@ class PreFlightResult:
             "reason": self.reason,
             "suggestion": self.suggestion,
             "triggering_joins": self.triggering_joins,
+            "certainty": self.certainty,
         }
 
 
@@ -1321,6 +1333,28 @@ def _one_side_facing_many(
     return hits
 
 
+def _known_table_names(
+    org: Datasource, rels: list[Relationship], ctx: "GuardContext | None" = None
+) -> set[str]:
+    """Every table the cardinality analysis can say anything about.
+
+    The declared tables PLUS the tables named on either end of a declared relationship — the
+    relationships are what the fan/chasm rules actually read, and a model can carry one for a
+    table it has not (yet) described in full."""
+    known = set(ctx.model_table_index if ctx is not None else _model_table_index(org))
+    for r in rels:
+        known.add(_bare(r.from_table))
+        known.add(_bare(r.to_table))
+    return known
+
+
+def _reads_any_table(
+    agg: "exp.AggFunc", tables: set[str], scope: dict[str, str]
+) -> bool:
+    """Does any column inside `agg` bind to one of `tables`?"""
+    return any(_resolve_col_table(col, scope) in tables for col in agg.find_all(exp.Column))
+
+
 def _many_side_facing_one(rels: list[Relationship], table: str, dim: str) -> bool:
     """Is `table` the MANY side of a join to dimension `dim` (the ONE side)?"""
     for r in rels:
@@ -1356,6 +1390,7 @@ def pre_flight_check(sql: str, org: Datasource,
     # Each arm is rendered back to text in the datasource's grammar before being re-read —
     # rendering generically would feed the analysis a statement in a grammar the arm was
     # never written in, the same mis-read this battery exists to prevent.
+    unresolved: Optional[PreFlightResult] = None
     for arm in _output_selects(tree):
         res = _preflight_select(arm, org, arm.sql(dialect=dialect), allow_rewrite=False, ctx=ctx)
         if res.risk and res.action == "refuse":
@@ -1367,7 +1402,21 @@ def pre_flight_check(sql: str, org: Datasource,
                 reason=res.reason,
                 suggestion=res.suggestion,
                 triggering_joins=res.triggering_joins,
+                certainty=res.certainty,
             )
+        if res.certainty == "uncertain" and unresolved is None:
+            unresolved = res
+    # An arm the analysis could not read makes the WHOLE set operation unproven — the arms are
+    # unioned into one answer, so an unexamined arm is an unexamined answer.
+    if unresolved is not None:
+        return PreFlightResult(
+            None,
+            "allow",
+            sql,
+            reason=unresolved.reason,
+            suggestion=unresolved.suggestion,
+            certainty="uncertain",
+        )
     return PreFlightResult(
         None, "allow", sql, reason="no fan/chasm or aggregation issue in any arm"
     )
@@ -1384,9 +1433,13 @@ def _preflight_select(tree: "exp.Select", org: Datasource, sql: str, allow_rewri
     rels = ctx.cardinality_index if ctx is not None else _cardinality_index(org)
     tables_in_scope = _tables_in_scope(tree)  # alias -> table
     table_set = set(tables_in_scope.values())
+    # Sources whose grain the analysis could NOT establish. An aggregate over one of these has
+    # unknown lineage, and the honest answer is "I could not tell".
+    opaque_sources = _unresolved_scope_sources(tree, _known_table_names(org, rels, ctx))
 
-    # aggregates: list of (table, is_aggregate)
-    agg_sources = _aggregate_source_tables(tree, tables_in_scope)
+    # aggregates: (source table, the aggregate node it appears in)
+    agg_pairs = _aggregate_source_tables(tree, tables_in_scope)
+    agg_sources = {t for t, _ in agg_pairs}
     has_raw_columns = _has_raw_non_grouped_columns(tree, tables_in_scope)
     has_aggregate = bool(agg_sources)
     no_aggregation = not has_aggregate
@@ -1420,8 +1473,18 @@ def _preflight_select(tree: "exp.Select", org: Datasource, sql: str, allow_rewri
                 triggering_joins=[f"{s} -> {shared}" for s in srcs],
             )
 
-    # FAN: an aggregate over a measure on the ONE side of a one-to-many in scope
-    for measure_table in agg_sources:
+    # FAN: an aggregate over a measure on the ONE side of a one-to-many in scope.
+    #
+    # Only aggregates a fan-out can actually inflate get a vote. MIN/MAX, a DISTINCT-qualified
+    # aggregate and a boolean fold are idempotent under row duplication, so the join cannot
+    # change their value and the query was already correct.
+    fan_aggs: dict[str, list["exp.AggFunc"]] = {}
+    for table, agg in agg_pairs:
+        if _is_fan_immune(agg):
+            continue
+        fan_aggs.setdefault(table, []).append(agg)
+
+    for measure_table in sorted(fan_aggs):  # sorted: one shape of query, one diagnosis
         others = table_set - {measure_table}
         fan_rels = _one_side_facing_many(rels, measure_table, others)
         if not fan_rels:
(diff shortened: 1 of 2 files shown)


sandeep-agami:
## Converting to draft — the loosening is unbounded in four shapes

Review found, and I independently reproduced through `pre_flight_check` against both branches, four queries that **refuse on base and are allowed here**. All four are genuinely inflated. All four are labelled `certainty="provable"`.

| query | base | this branch |
|---|---|---|
| `SUM(CASE WHEN order_items.quantity > 0 THEN orders.total_amount ELSE 0 END)` | `refuse` | **allow / provable** |
| `STRING_AGG(orders.status, ',' ORDER BY order_items.id)` | `refuse` | **allow / provable** |
| `WITH oi AS (SELECT order_id, product_id, SUM(quantity) … GROUP BY order_id, product_id) … JOIN oi ON oi.order_id = orders.id` | `refuse` | **allow / provable** |
| `WITH oi AS (SELECT DISTINCT order_id, product_id …) … JOIN oi ON oi.order_id = orders.id` | `refuse` | **allow / provable** |

### Why each is wrong

**The first two — the co-factor rule is too broad.** `_reads_any_table` is a bare `find_all(exp.Column)` over the whole aggregate node, so *any* syntactic occurrence of a many-side column exempts the aggregate. The rule's justification holds when the many-side column is a **factor of the value** (`quantity * total_amount`). It does not hold when the column is a boolean selector or an ordering key: in the `CASE` query the aggregated value is `orders.total_amount`, and an order with three qualifying line items contributes it three times. Conditional aggregation over a fan is how every *"revenue of orders containing product X"* query is written — this is the shape most likely to reach a customer.

**The second two — a grain-changing CTE is assumed to sit at the join key's grain, and never checked against it.** `oi` is one row per *(order, product)*, not per order, so the join still fans. Nothing compares the CTE's `GROUP BY` / `DISTINCT` keys to the columns it is joined on. This is a strictly easier bypass than the `WITH o AS (SELECT * FROM orders)` shape the change set out to close, because it survives the new scope walk **by design**.

### Two more, same root

`SELECT SUM(d.total_amount) FROM (SELECT * FROM orders) d JOIN order_items …` — the derived-table spelling of the under-detection case — is `allow` / `provable`. Deferring derived-table lineage is a legitimate scope call; labelling it `provable` is not, since that is the field the tiered policy will key enforcement off.

And **`certainty` reaches no surface**: no caller reads it, `sm prepare` drops the reason when `risk` is `None`, and both receipt builders gate on `if pre_flight.risk:`. So the `uncertain` path — the whole answer to "never a silent allow" — is behaviourally identical to base's silent allow. The success criterion passes because it asserts on the field, not on what happens.

### My error, not the reviewer's find

The non-regression criterion I wrote only guarded `allow → refuse`. Every one of these is `refuse → allow`, which nothing in the spec covered and which the corpus would not catch. Loosening fixes need the reverse assertion, and the spec asked for it in only one direction.

Next commit adds that reverse assertion first, then narrows the two rules.

sandeep-agami:

**Spec:** ACE-083 (trap classification)

Treat this as a reference implementation, not a pending review. Leaving it open falsely signals that a review is owed. The work is being rebuilt against `main` under ACE-083.

Nothing is lost. This PR's code is preserved at the immutable tag `governance-reference-ace-083` (`2020d9f`), and the base branch at `governance-branch-reference` (`76ed6cb`). `agami-governance-branch` is being deleted, so read reference code from the tags, never from a branch name.

PR #161, opened 2026-07-29
ACE-040: make the DB-required sentinel actually fail, and close five corpus gaps

sandeep-agami:
**Spec: ACE-040**

Test and CI infrastructure only — no file under `packages/agami-core/src/` changes.

## The sentinel was defeated by two independent paths

`AGAMI_IT_PG_REQUIRED` exists so the Postgres integration job hard-fails rather than skips when a database is expected. It did not work. With the sentinel set and drivers absent:

```
EXIT=0
7 skipped in 0.65s
SKIPPED [1] test_safety_corpus.py:18: could not import 'mcp'
SKIPPED [1] test_role_floor_pg.py:29: could not import 'psycopg2'
SKIPPED [5] test_role_floor_pg.py:36: could not import 'psycopg2'
```

Two causes, not one: the driver import in `pg_admin`, **and** four module-level `importorskip` calls that take the whole corpus module out before the sentinel is ever consulted. The DB-unreachable path already behaved correctly (exit 1); only the dependency path was open.

`tests/e2e/itdeps.py` adds `importorfail` — `importorskip` when the sentinel is unset, hard failure when set — wired into both. `tests/e2e/test_it_sentinel.py` runs the job's own invocation in a subprocess with a dependency blocked and asserts the exit code both ways. Verified failing on base by reverting only the two call-site edits: both sentinel arms failed, both "still skips" arms passed.

Two things worth recording from building it:
- A blocker raising plain `ImportError` is **not** caught by this pytest's `importorskip`, which defaults `exc_type` to `ModuleNotFoundError` — an early version made base look stricter than it is.
- A bare module-scope `pytest.skip` is a collection **error**, so a first cut would have turned the DB-free job red whenever `mcp` / `starlette` / `sqlglot` / `pydantic` were absent. Hence the explicit `module_level` flag.

## Corpus gaps

Nine cases added — `pg_read_file`, `lo_import`, `dblink`, `dblink_exec`, `SELECT …; COMMIT`, `SET ROLE postgres`, a line comment hiding a statement separator, and CTE-hidden `DELETE` / `INSERT` — plus a cross-join runaway asserting `"bounded"`.

**These are pure coverage gaps and passed immediately. They caught nothing.** Every one was probed first on both paths and already refused with `permission`, identically on SQLite and Postgres.

## A trap for future authors

The CI job selects with `-k "db_path or role"`, so **a case label containing `role` silently pulls file-path tests into the DB job** — an early label did exactly that, 108 → 110. Renamed, and the coupling is now documented on `Case.id` where an author will see it.

## Verification

Postgres job: **88 → 108 passed**. Full `tests/e2e/`: **230 passed**. DB-free suite: **2280 → 2306**.

`dev.py check`: **2306 passed, 115 skipped, 2 xfailed**; ruff + gitleaks clean; no vendored-lib drift.

One unexplained transient: in six job runs, one reported `107 passed, 1 error`; four consecutive re-runs were clean at 108. It followed a working-tree rewrite mid-session, so most likely stale bytecode from local tooling — but the traceback was not captured, so flakiness in the role fixtures (which drop and recreate a global role per test, already marked serial-only) is not ruled out.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -17,6 +17,7 @@ if str(REPO_ROOT / "tests") not in sys.path:
     sys.path.insert(0, str(REPO_ROOT / "tests"))
 
 import harness  # noqa: E402  (tests/e2e is on sys.path during collection)
+import itdeps  # noqa: E402  (same directory; skip-vs-fail gating for the DB-backed tests)
 
 BASE = "https://demo.example.com"
 
@@ -89,17 +90,22 @@ def pg_admin():
     BUT when AGAMI_IT_PG_REQUIRED is set (the integration-pg CI job sets it), an unavailable DB FAILS
     instead of skips — this job carries the ONLY proof of the role-floor + file/db parity + DB-served
     model, and pytest exits 0 when everything skips, so a service race / env rename / driver hiccup
-    would otherwise turn the F9 done-bar gate green while proving nothing."""
-    psycopg2 = pytest.importorskip("psycopg2")
+    would otherwise turn the F9 done-bar gate green while proving nothing.
+
+    The MISSING-DRIVER case runs through the same sentinel (`itdeps.importorfail`, not
+    `pytest.importorskip`): an absent psycopg2 is exactly the "proved nothing" outcome the sentinel
+    exists to catch, and importorskip would have skipped past it before the sentinel was consulted."""
+    psycopg2 = itdeps.importorfail("psycopg2")
     # In the required job, a missing DB is a hard failure — an all-skip must NOT pass as green.
-    unavailable = pytest.fail if os.environ.get("AGAMI_IT_PG_REQUIRED") else pytest.skip
     if not os.environ.get("AGAMI_IT_PG_PASSWORD"):
-        unavailable("set AGAMI_IT_PG_PASSWORD to run the Postgres safety-corpus / role-floor tests")
+        itdeps.unavailable(
+            "set AGAMI_IT_PG_PASSWORD to run the Postgres safety-corpus / role-floor tests"
+        )
     sc = pg_super_creds()
     try:
         conn = psycopg2.connect(connect_timeout=10, **sc)
     except Exception as exc:  # unreachable DB → skip locally, FAIL in the required CI job
-        unavailable(f"no reachable Postgres ({exc})")
+        itdeps.unavailable(f"no reachable Postgres ({exc})")
     conn.autocommit = True
     try:
         yield psycopg2, conn, sc
--- a/tests/e2e/itdeps.py
+++ b/tests/e2e/itdeps.py
@@ -1,0 +1,49 @@
+"""Dependency gating for the env-gated Postgres integration tests — the one place that decides
+whether a missing prerequisite is a SKIP or a FAILURE.
+
+`AGAMI_IT_PG_REQUIRED` exists so the DB-backed job cannot pass while proving nothing: that job
+carries the only evidence for the Postgres-served safety corpus (file/db parity) and the read-only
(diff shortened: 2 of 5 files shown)


sandeep-agami:
## Converting to draft — there is a fourth defeat path, and it is the one that matters

The fix wraps five **dependency** sites. But the job's real precondition is *"the DB-backed tests ran"*, and nothing asserts that. Reviewer demonstrated it: rename `test_safety_corpus_db_path` → `test_safety_corpus_over_the_database` (an ordinary refactor), leave the sentinel set and Postgres up, run the exact CI command:

```
6 passed, 214 deselected in 0.87s
```

**Exit 0. Green.** The entire Postgres-served corpus — the file/db parity proof, the whole reason the job exists — is gone, and the sentinel says nothing, because no import failed. An empty `DB_PATH_CASES` does the same thing (`got empty parameter set` → exit 0), and so will any future skip source not routed through `itdeps`.

So the change fixes the defeat path that was reported and leaves a wider one open. Wrapping call sites one at a time is the wrong shape — the floor has to be asserted **once, per session, derived from the corpus** so it cannot drift:

```python
def pytest_sessionfinish(session, exitstatus):
    if not itdeps.db_required():
        return
    expected = 2 * len([c for c in CASES if c.runs_on("PostgreSQL")]) + 6   # 2 surfaces + role floor
    if _DB_PROOF_PASSED < expected:
        session.exitstatus = 1
```

That closes the rename gap, the empty-parametrize gap, a `-k` expression that stops matching, and any future `skipif` — in one hook rather than one patch per call site as someone remembers.

## `session:tcl-commit` proves nothing about TCL

Probed `_read_only_reason` directly: `SELECT id FROM orders; COMMIT` returns **`multiple statements are not allowed — send one SELECT`**. Rung 2 fires before rung 4, so `COMMIT`'s deny-list membership is never consulted — delete `COMMIT` from the deny list and the case still passes. It is also a duplicate of the existing `integrity:multi-statement`, and the note above it describes a control the case never reaches. `SET ROLE postgres` is the same class, milder: it is caught by the statement-opening rung, not the `SET` deny-keyword entry.

Worth stating plainly in the corpus that rungs 2 and 3 between them make the TCL deny-list entries unreachable by any single statement. That is useful for the next author, and better than a case implying coverage it does not have.

**The other seven land exactly where claimed** — verified the same way: `pg_read_file` / `lo_import` / `dblink` / `dblink_exec` hit the dangerous-function rung by name; `line-comment-hides-separator` hits `DROP` *after* neutralization, precisely the claim; the two CTE-hidden cases hit `DELETE` / `INSERT` inside a `WITH` that passed the opening check. And `cross-join-runaway` correctly asserts `bounded`.

## Also

The `-k` collision note sits on the `Case.id` property, 155 lines from where anyone appends a case — and it is advisory. Make it mechanical: a test asserting no `Case.id` contains `role` or `db_path`. A comment a machine checks is worth ten that it doesn't.

Confirmed clean: no runtime source changed (`git diff --name-only … -- packages/ plugins/ deploy/` → 0 files), the DB-free path still skips rather than erroring, the subprocess test is not vacuous, and there is no live `-k` collision today.

sandeep-agami:

**Spec:** ACE-040 (DB-required sentinel)

Treat this as a reference implementation, not a pending review. Leaving it open falsely signals that a review is owed. The work is being rebuilt against `main` under ACE-040.

Nothing is lost. This PR's code is preserved at the immutable tag `governance-reference-ace-040` (`ba62f85`), and the base branch at `governance-branch-reference` (`76ed6cb`). `agami-governance-branch` is being deleted, so read reference code from the tags, never from a branch name.

PR #162, opened 2026-07-29
ACE-082: prove the vetted statement is one the engine accepts — MySQL, SQL Server, DuckDB

sandeep-agami:
Spec: ACE-082

## Summary

ACE-079 made the guard read every statement in the datasource's own dialect, and proved the **verdict** on all eleven supported engines. But that proof runs a spy executor against **no database** and re-parses the vetted statement with sqlglot — it is sqlglot asserting about sqlglot. The other half of REQ-014 fidelity, *the statement the guard vetted is a statement the engine actually accepts*, was proven only on **SQLite and PostgreSQL** — both `"`-quoting, i.e. precisely the family where ACE-079's defect class **cannot manifest**. The fix was verified only where the bug could not occur.

This runs the same corpus, through the same tool, against **real MySQL 8, real SQL Server 2022, and in-process DuckDB**. With SQLite and Postgres that is five engines executing, covering **all three identifier-quoting families** — backtick, bracket, double-quote — which is the axis the defect turns on. Each of the six that remain verdict-only shares a quoting family with one that executes.

The divergence this exists to catch, reproduced on the pre-ACE-079 revision (`8d63ccb`):

```
SELECT TOP 5 [id], [amount] FROM [orders]
  BEFORE (generic parse, errors ignored) -> tables=[]        cols=['TOP']
  AFTER  (tsql, the declared engine)     -> tables=['orders'] cols=['amount','id']
```

**No runtime source changed.** `git diff origin/agami-governance-branch -- packages/agami-core/src/` is empty, and two engines were dropped rather than break that rule (below).

## Changes

**The seam** — the spec assumed a per-engine fixture seam existed; it did not. SQLite was seeded by `harness.seed_sqlite`, Postgres by a *separate* inline seeder in `conftest.db_safety_env`, each with its own type map and placeholder style. Both now go through one `EngineFixture` + `seed_schema`, so a DDL-spelling difference lives in a table and adding an engine is a row, not a second seeder that can drift.

**New cases**, pinned via the existing `Case.engines`:
- a backtick-quoted governed query is allowed **and executes** on MySQL, while a backtick-quoted undeclared table is refused;
- `SELECT TOP n [col] FROM [tbl]` is scoped correctly on SQL Server;
- on MySQL a double-quoted statement is `unscopable_sql`, because there `"` is a **string literal**. The generic reading of that same text is a perfectly scopable query over `orders` — so this is the positive proof that reading in the engine's own grammar changes the **verdict**, not merely the parse. The two `"`-quoting cases that ran everywhere are now pinned to the engines where `"` actually quotes an identifier.

**CI** — three separate jobs, no `needs:` on `integration-pg` or on each other, so one flaky container cannot move the F9 done-bar. `integration-pg`'s `-k "db_path or role"` selector is unchanged and matches **zero** new tests — now asserted by a test rather than assumed. No secrets, so they run on fork PRs.

**Two pre-existing tests fixed** (both latent, both surfaced by this work, neither weakened — each now checks strictly more):
- `test_no_false_refusal_on_valid_sql.py` parsed every valid case under a hardcoded `sqlite` dialect despite its docstring claiming per-engine parsing. Now measures (case, engine) pairs: **15 → 23 checks**, with all 15 originals retained.
- `test_resolve_guard_model.py` loaded `harness.py` via `spec_from_file_location` **without** registering it in `sys.modules`, so adding any class that reads module globals at creation time broke it at import. Fixed to the documented idiom.

## Decisions

- **CI does not depend on cloud credentials.** Tier 3b (BigQuery dry-run, Snowflake `EXPLAIN`) is deferred. No credentials exist; the marginal evidence is low (Snowflake and Redshift are `"`-quoting, the family Postgres already covers); a public repo withholds secrets from fork PRs, making such a job structurally unrunnable for outside contributors; and it is a standing maintenance tax. If ever wanted, the right shape is a non-required `workflow_dispatch`/`schedule` job. **Residual risk: real BigQuery and Snowflake grammar stays unproven by execution.**
- **BigQuery emulator and local Spark were dropped as unreachable, not skipped by preference.** `_run_bigquery` builds `bigquery.Client(project=…)` with no `client_options`/`api_endpoint`, so it cannot be pointed at an emulator; `_run_databricks` speaks to a workspace, not a local `SparkSession`. Reaching either needs a runtime source change this spec forbids, and going through a stub would assert exactly what ACE-079 already asserts. (The emulator is also SQLite-backed, so it would have proven parse/accept, not BigQuery semantics.)
- **One job per engine.** Free (parallel, no `needs:`) and buys failure isolation: SQL Server is the heavy, slow-booting container and should not be able to mask the backtick evidence.
- **"Unavailable" includes a missing driver.** Review reproduced a hole where `importorskip` ran before the `AGAMI_IT_<ENGINE>_REQUIRED` check, so a job whose driver failed to install skipped everything and exited 0 — green, proving nothing, in the job that owns that engine's only evidence. Closed and verified in both directions.

## Checklist

- [x] `dev.py check` green — 2291 passed, 359 skipped, 2 xfailed
- [x] Zero files changed under `packages/agami-core/src/`
- [x] Live runs: MySQL 92 · SQL Server 90 · DuckDB 82 · Postgres (db_path + role floor) 88
- [x] Skip-vs-fail verified both directions, for an unreachable server **and** a missing driver
- [x] `integration-pg` selector unchanged; independence asserted by a test
- [x] Per-engine coverage tier stated in one place, including what is *not* proven
- [x] No test weakened or deleted; the two edited tests check strictly more
- [x] gitleaks clean; no customer names, internal hosts, or real credentials (CI literals are labelled non-secret sentinels for ephemeral localhost services)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,136 @@ jobs:
           pytest tests/e2e/test_safety_corpus.py tests/e2e/test_role_floor_pg.py
           -q -k "db_path or role"
 
+  # ── ACE-082: per-engine EXECUTION evidence ───────────────────────────────────────────────────
+  # Deliberately SEPARATE jobs, with no `needs:` on integration-pg and none between themselves, so
+  # one flaky engine container cannot turn the Postgres role-floor result (the F9 done-bar) green
+  # or red. Each job selects only its own engine, and sets AGAMI_IT_<ENGINE>_REQUIRED so an
+  # unreachable server FAILS rather than skipping — pytest exits 0 when everything skips, and a
+  # silently-skipped job would claim per-engine coverage while proving nothing.
+  # These need no secrets, so they run on fork PRs too.
+
+  integration-mysql:
+    # The highest-value engine here: backticks are MySQL's NATIVE identifier quote, so this is
+    # where a wrongly-dialected parse and a real execution can genuinely disagree.
+    name: integration (MySQL — dialect execution matrix)
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql:8
+        env:
+          MYSQL_DATABASE: shop
+          MYSQL_USER: agami_test
+          # Ephemeral localhost CI service on a throwaway container. MySQL has no `trust` auth
+          # equivalent, so a literal is unavoidable; this is a non-secret sentinel, not a credential.
+          MYSQL_PASSWORD: ci_local_only
+          MYSQL_ROOT_PASSWORD: ci_local_only_root
+        ports:
+          - 3306:3306
+        # `mysqladmin ping` exits 0 once the server answers, even for an unauthenticated ping —
+        # which is exactly the readiness signal we want.
+        options: >-
+          --health-cmd "mysqladmin ping"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 20
+    env:
+      AGAMI_IT_MYSQL_HOST: 127.0.0.1
+      AGAMI_IT_MYSQL_PORT: "3306"
+      AGAMI_IT_MYSQL_USER: agami_test
+      AGAMI_IT_MYSQL_PASSWORD: ci_local_only
+      AGAMI_IT_MYSQL_DB: shop
+      AGAMI_IT_MYSQL_REQUIRED: "1"
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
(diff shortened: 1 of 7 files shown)


sandeep-agami:

**Spec:** ACE-082 (dialect execution matrix)

Treat this as a reference implementation, not a pending review. Leaving it open falsely signals that a review is owed. The work is being rebuilt against `main` under ACE-082.

Nothing is lost. This PR's code is preserved at the immutable tag `governance-reference-ace-082` (`5adf1b5`), and the base branch at `governance-branch-reference` (`76ed6cb`). `agami-governance-branch` is being deleted, so read reference code from the tags, never from a branch name.

PR #164, opened 2026-07-29
ci: enforce the coverage floor (85%) instead of only reporting it

ashwin-agami:
**Spec: OCR-025** — the *"make the CI gate actually block"* line of work.

## What was wrong

The `pytest --cov` job printed a coverage report and **failed on nothing**. `--cov-report=term-missing` with no `--cov-fail-under` anywhere in the workflow.

**Coverage could regress to zero and CI would stay green.** OCR-024's spec states a `pytest>=80%` floor — it has never existed in the workflow.

## The number, measured rather than assumed

| Python | Coverage | Tests |
|---|---|---|
| 3.13 | **86%** | 1584 passed, 1 skipped |
| 3.10 | **86%** | identical |

Consistent across the matrix, so the floor goes in one point below at **85%** — enforced immediately, no change to the suite, no risk to any matrix leg.

Verified locally with the exact CI command:

```
Required test coverage of 85% reached. Total coverage: 85.66%
1584 passed, 1 skipped in 55.18s
```

**80% was the obvious number and is the wrong one** — it would license a six-point regression from where the suite actually is.

## Deliberately not in this PR

Flipping `ruff format --check` off `continue-on-error`.

The tree needs **119 files** reformatted — the comment saying ~74 is stale — across roughly **11,600 lines**. And **23 of those files are touched by the 13 currently-open pull requests**.

A mechanical reformat now forces every one of those PRs to rebase through a diff where every line moved, which is the most painful conflict to resolve and the easiest to resolve wrongly.

Do it when the queue is short, in its own commit, and add that commit to `.git-blame-ignore-revs` so it doesn't bury the history.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,10 +38,11 @@ jobs:
         run: uvx ruff@0.15.19 format --check plugins packages tests
         continue-on-error: true
 
-      # Coverage is ENFORCED, not just reported. Measured at 86% on 2026-07-29 (identical on 3.10 and
-      # 3.13, 1584 tests), so the floor sits one point below with no change to the suite. Before this
-      # the job printed a coverage report and failed on nothing — OCR-024 claimed a >=80% floor that
-      # was never in the workflow, so coverage could regress to zero and CI stayed green.
+      # Coverage is ENFORCED, not just reported. Measured 2026-07-29 at 85.66% on every leg of this
+      # matrix — 3.10, 3.11 and 3.12 alike, 1584 tests — so the floor sits just below it and no matrix
+      # leg is closer to the line than any other. Before this the job printed a coverage report and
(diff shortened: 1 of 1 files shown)


ashwin-agami:
Good catch — that was a real inaccuracy, not a nit. The comment cited **3.13** because that's my local default, but this matrix runs **3.10–3.12**. A rationale naming a version the workflow never builds on is worse than one naming no version at all.

Rather than reword it away, I measured the two legs I hadn't:

| Python | Coverage | Tests |
|---|---|---|
| 3.10 | 85.66% | 1584 passed, 1 skipped |
| 3.11 | 85.66% | 1584 passed, 1 skipped |
| 3.12 | 85.66% | 1584 passed, 1 skipped |

Identical across the whole matrix, which is worth recording in itself — **no leg sits closer to the floor than any other**, so the 85% floor can't be tripped by one version drifting alone.

The comment now cites the matrix it actually belongs to.

PR #165, opened 2026-07-29
Per-session identity on the access token + fix the guard model's org lookup

vishalkalbi27:
Two changes to the served path, both found by running a real multi-tenant deployment.

## 1. `fix(ACE-051)`: the safety pass looked the model up under `'local'`

`execute_sql._resolve_guard_model` called `load_datasource(store, profile)`, taking the `org_id='local'`
default, while its own docstring says it mirrors `tools._load_org` — which passes `org_id=_current_org_id()`.

That agreed once. In 0.4.x `model_deploy._default_org()` was also `AGAMI_ORG_ID or 'local'`, so rows were
written under `'local'` and read under `'local'`. F14/F15 moved the write side onto the deployment's resolved id
without moving this reader, producing exactly the failure `_default_org`'s docstring warns about: *"the model
is written under one org and read under another and the server sees no model."*

On a multi-tenant server every tenant's rows live under its own id, so the `'local'` lookup missed for **all**
tenants and the fail-closed rule refused **every** query with `model_unavailable`. Single-tenant deploys were
spared only because the disk fallback quietly rescued them — which is why it survived a release unnoticed.

The regression test seeds a model under a named tenant and asserts the guard resolves it with the artifacts
dir pointed at nothing, so the disk fallback cannot mask it again.

## 2. `feat(auth)`: a per-session identifier

A `Principal` carried one field — `subject` — so two clients of the same login were indistinguishable.
Anything remembered per-caller was therefore shared, last write wins. Observed live: one window selected
tenant A, a second selected tenant B, and the first window — still displaying A — read B's data. No error;
the resolver did exactly what it was told.

Access tokens now carry `sid`, and `Principal` gains an optional `session_id`.

**Why the refresh family, not a `jti`:** a `jti` changes on every refresh, so anything keyed on it silently
resets each hour — the same class of bug being fixed. The family survives rotation (`rotate` carries it into
the successor row; `overwrite` never rewrites the column), so it identifies one client authorization for its
whole life.

**Why hashed** (`sha256(family)[:16]`): the raw family is the key reuse detection burns on (`WHERE family = ?`)
and should not leave the server.

Threading it needed one real change — the family was previously unrecoverable. `_issue_refresh_token`
computed `family or token_urlsafe(16)` inline inside the INSERT tuple and returned only the token, so on the
authorization_code path (the one that *mints* a lineage) no frame ever learned it. It now returns the pair.

`sid` is optional at every step and deliberately absent from the `require` list: presence auth, hand-minted
bearers and username/password all yield `session_id=None`, and a malformed claim degrades to None rather than
costing a valid caller their request. Reads use `getattr` because `AuthProvider` is `@runtime_checkable` —
method presence only — so a third-party principal may expose nothing but `.subject`.

`sid` is an **identity, not a liveness signal**: reuse detection can burn a family while tokens carrying its
`sid` are unexpired, and `validate_token` does no DB lookup. Documented at the mint site.

Nothing consumes it yet — this is the seam. The consumer side lands in agami-hosted after a release.

## Blast radius

Local flows are insulated twice over: the changed branch is gated on `_hosted()` (false with no
`AGAMI_DB_URL`), and the vendored plugin slice ships no `store`/`model_store`/`tools`, so it cannot run that
branch at all. stdio imports neither `oauth_server` nor `mcp_http`. Only the hosted HTTP server is affected.

## Verification

- **1597 passed, 1 skipped, 0 failed**; ruff + gitleaks clean
- Reverting any of the three pieces — the guard's `org_id`, the contextvar set, or the `sid` on the token
  body — fails a test that names the behaviour
- No vendored module drift beyond the intended `execute_sql.py` re-sync

⚠ **Run the suite with `AGAMI_ARTIFACTS_DIR` isolated.** Without it, 32 unrelated tests fail on any machine
with a real agami deployment: core resolves `~/.config/agami/path` and `resolved_org_id()` returns that org
instead of `'local'`, while fixtures seed under `'local'`. CI is green only because a runner has no deployment.
Separate latent issue, worth its own fix (an autouse fixture in `tests/conftest.py`).

### DIFF
--- a/packages/agami-core/src/execute_sql.py
+++ b/packages/agami-core/src/execute_sql.py
@@ -883,11 +883,19 @@ def _resolve_guard_model(profile: str):
         try:
             from model_store import load_datasource as _load_db
             from store import Store
+            from tools import _current_org_id
 
             store = Store.from_env()
             if store is not None:
                 try:
-                    org = _load_db(store, profile)
+                    # Scoped to the REQUEST's org, exactly as `tools._load_org` does. Letting this
+                    # default to 'local' looked right while `model_deploy._default_org()` was also
+                    # `AGAMI_ORG_ID or "local"`, but F14/F15 moved the write side onto the deployment's
+                    # resolved id — so rows land under that id and a 'local' read finds nothing. On a
+                    # multi-tenant server it finds nothing for EVERY tenant, and the fail-closed rule
+                    # below then refuses every query. `_default_org`'s own docstring states the contract
+                    # this restores: the write and read org "MUST agree".
+                    org = _load_db(store, profile, org_id=_current_org_id())
                 finally:
                     store.close()
                 if org is not None:
--- a/packages/agami-core/src/mcp_http.py
+++ b/packages/agami-core/src/mcp_http.py
@@ -63,6 +63,21 @@ _log = logging.getLogger(__name__)
 # only receives (name, arguments), and a contextvar set in the BaseHTTPMiddleware wouldn't reach it.
 _actor_ctx: ContextVar[str | None] = ContextVar("agami_tool_actor", default=None)
 
+# The CLIENT AUTHORIZATION behind the in-flight tool call — what tells two windows under one login apart.
+# Set alongside `_actor_ctx` and for the same reason. None whenever the caller's auth has no session
+# notion (presence auth, a hand-minted bearer): consumers must read that as "no session", not an error.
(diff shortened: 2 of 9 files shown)


ashwin-agami (commented):
Is this the same way that it would work if we implement google/microsoft login or implement custom login for other OIDC providers?

vishalkalbi27:
> Is this the same way that it would work if we implement google/microsoft login or implement custom login for other OIDC providers?

Yes it will work the same way

PR #180, opened 2026-08-02
ACE-042: stop injecting declared filters — delete the last authoring path in the guard

sandeep-agami:
Spec: ACE-042

## Summary

`runtime.apply_default_filters` ANDed each in-scope table's declared `default_filters` into the statement before execution. Three things were wrong with that, in increasing order of severity.

**It was misclassified.** `default_filters` are business logic (`is_deleted = false`, `status != 'test'`), a statement about what the org *means* by a table, not a security boundary. Getting one wrong produces a wrong answer, not a leak.

**It injected.** Agami never authors or alters SQL. The one carve-out that permitted this transform rested on the misclassification above, so with the classification corrected the carve-out has no grounds.

**And it was broken.** `_tables_in_scope` is `tree.find_all(exp.Table)`, which descends into CTEs, so an alias bound *inside* a CTE was collected and its filter ANDed onto the **outer** `WHERE`, where that alias does not exist. Reproduced verbatim against `main`:

```
in:  WITH recent AS (SELECT id, total FROM orders o WHERE o.created_at > '2026-01-01')
     SELECT SUM(total) FROM recent
out: WITH recent AS (SELECT id, total FROM orders AS o WHERE o.created_at > '2026-01-01')
     SELECT SUM(total) FROM recent WHERE (o.is_deleted = FALSE)
applied: ['o.is_deleted = false']
```

The database rejects that statement and the receipt says the filter was applied: our own edit manufacturing a database failure. `test_the_cte_case_reaches_the_driver_unchanged` pins it, and all five new tests fail against pre-change code.

This PR is pure subtraction. Reporting which filters a statement applied is ACE-099, and this deliberately ships before it: between the two, a declared filter is neither applied nor reported. That window is accepted (the injection was already undependable, and keeping it means shipping an injector that manufactures failures) on condition it is **announced**, which is what the notices below do.

## Changes

- **Deleted `runtime.apply_default_filters`** outright, with its `__all__` entry and the ACE-045 `ctx.tree.copy()` that existed only because it mutated its tree. Not left as a `-> list[str]` reporter: that is ACE-099's job a wave early, built on the same `find_all` scoping ACE-099 exists to replace.
- **Removed both call sites** — `execute_sql.py::_model_safety` (in both vendored copies) and `semantic_model/cli.py::cmd_prepare`. The second was not named in the spec; grounding found it. `sm prepare` keeps its fan/chasm pre-flight and drops its `applied_filters` key rather than emit a permanently empty list, which would read as "we checked, none applied".
- **Stated the window** in the `execute_sql` tool description, `SERVER_INSTRUCTIONS`, `Table.default_filters`, the `_model_safety` and `cmd_prepare` docstrings, both shipped skills, `docs/format-spec.md`, and the model-explorer GUI (where the field is actually authored, and where it was labelled "Always-on filters"). Each notice states the behavior; the `ACE-099` deletion anchor lives in the adjacent comment, matching this repo's precedent that spec ids appear in comments and docstrings but never in shipped strings.
- **Untouched, by contract:** `_tables_in_scope` (ACE-099 inherits its naive walk deliberately; it has three other callers), `_drop_fanout_joins` and byte-identity (ACE-093), `assemble_receipt` / `receipt["default_filters_applied"]` / `sm receipt --applied-filters` (ACE-088, ACE-099).

## Tests changed — each one a deliberate contract change

Recorded in the spec's `## Decisions` as well.

- **Deleted:** the three `apply_default_filters` tests in `test_semantic_model_runtime.py` and their `_filter_org` helper. They assert the injection being removed; `tests/test_ace042_no_filter_injection.py` replaces them.
- **Rewritten:** `test_guard_context.py` (the call leaves the no-re-parse and parity batteries; the property survives for the remaining five guards), `test_prepare_allow_applies_default_filters` → `test_prepare_does_not_apply_default_filters`, `test_prepare_auto_rewrites_fan_trap`'s filter assertion.
- **Retargeted:** two synthetic stderr fixtures move from `[agami] applied default_filters: …` to `[agami] auto-corrected …`, which still occurs. ACE-093 deletes that notice too and will have to move them again; accepted over quoting a string the code can no longer emit.
- **Demoted, honestly:** `test_ace039_error_detail.py`'s end-to-end relay test. With the notice gone the child writes only the sanitized sentence, so relay and reconstruct agree and a mutation to pure relay leaves the file green. Its docstring now says it is a smoke test and points at the discriminating version in `test_ace035_read_only_refusal.py`.

## Smoke: the defect, against a real database

The unit tests assert the statement is *unchanged*; they never execute it. This runs the same three cases through the real executor against the repo's own SQLite shop fixture (10 orders, 2 of them `status = 'test'`), on `main` and on this branch, with `orders` declaring `default_filters: ["{alias}.status != 'test'"]`.

**1. The CTE case.**

```
main      [agami] applied default_filters: ["o.status != 'test'"]
          The statement referenced a column this database does not have.   exit=7
branch    n=9                                                              exit=0
```

The manufactured database failure, end to end: `o` is bound only inside the CTE, the filter landed on the outer `WHERE`, the driver rejected it, and the notice claimed it had been applied.

**2. Asking about the rows the filter excludes — worse than "never refused".**

```
SELECT COUNT(orders.id) AS n FROM orders WHERE orders.status = 'test'
main      n=0
branch    n=2
```

`main` ANDed `status != 'test'` onto a query whose `WHERE` was `status = 'test'`, making a contradiction that returns zero. "How many test orders are there?" answered **zero** — not a refusal, not an error, a confidently wrong number with a receipt asserting the filter was applied. SC-3 is worded as "executes and is never refused", which undersells the old behaviour: the failure mode was a silent wrong answer, the exact class this layer exists to prevent.

**3. The window, quantified.** `SELECT COUNT(orders.id) FROM orders` returns **8** on `main` and **10** on the branch, against 10 orders of which 2 are `test`. Exactly the disclosed consequence.

**4. `sm prepare`.** `main` returns re-serialised SQL plus `applied_filters: ["orders.status != 'test'"]`; the branch returns the caller's statement verbatim, no `applied_filters` key.

## Checklist

- [x] `uv run dev.py check` green (ruff · pytest · gitleaks · vendored-drift) — 2416 passed, 4 skipped; patch coverage 100%. CI green on py3.10 / 3.11 / 3.12.
- [x] Linked to a spec above
- [x] Tests added/updated — five new, seven changed, each recorded above
- [x] Docs/decisions recorded — spec `## Decisions` (agami-sdlc `3b10409`, `80b4c37`) and this body
- [x] No secrets committed

## For the reviewer

Two things worth your eyes:

1. **The reclassification** (row-scoping is business logic, not a disclosure control) is what licenses this deletion. The security pass pressure-tested it and it held: the `:param` tenancy form was **never** injected on any shipped path (the body skipped unresolved bind markers and no caller ever passed `params`), per-principal row visibility does not exist in free core, and all shipped sample tables declare `default_filters: []`. The residual is a *literal* tenancy pin, which is a per-deployment constant identical for every caller and so never distinguished principals.
2. **The Copilot round found one real thing.** An earlier commit stripped spec ids from the `tools.py` strings because they reach the model, then parked the `ACE-099` deletion anchor in an HTML comment on two SKILL.md files and `docs/format-spec.md`. A skill file **is** the prompt, loaded verbatim, so that comment reached the model exactly like the prose beside it — the same leak, in comment syntax. Removed, and `test_no_shipped_markdown_surface_carries_a_spec_id` now holds the line (`main` had zero across those trees; these three were the only ones).

3. **ACE-043 is `approved` and now contradicts this.** Its In scope still says to "AND each arm's in-scope `default_filters` into that arm's `WHERE`", which re-adds what this deletes. ACE-042 lands first so it is not a block, and this PR leaves no set-op scaffolding for it, but ACE-043 needs rewriting into the reporting spec its neighbours describe.


sandeep-agami:
Addressed the Copilot review in `12a0960`. All three comments were the same finding, and it was a real one.

**The inconsistency was mine.** An earlier commit in this PR stripped spec ids out of the `tools.py` tool description and `SERVER_INSTRUCTIONS` precisely because those strings reach the model — then parked the `ACE-099` deletion anchor in `<!-- ACE-042 -> ACE-099 -->` on two SKILL.md files and `docs/format-spec.md`, on the reasoning that markdown has no `#` comment available. That reasoning does not survive contact with how a skill is loaded: a SKILL.md **is** the prompt, read verbatim, so the HTML comment reached the model exactly like the prose beside it. Same leak, wearing comment syntax.

All three comments are removed. The behavioural statement each one sat next to is untouched, which is all the spec's criterion actually asks for.

**Made it a standing guard rather than a one-time fix.** `origin/main` carries zero spec ids across `plugins/agami/skills/`, `plugins/agami/shared/` and `docs/` — these three were the only ones in the tree. `test_no_shipped_markdown_surface_carries_a_spec_id` now fails the build on any future one, so this does not depend on the next author remembering.

**On what the anchor was for.** It existed so the follow-up spec could grep for the notices it has to delete. Losing it on those three surfaces is a real cost, so the spec's `## Decisions` now names each notice by location instead — the two `default_filters` paragraphs in `agami-query/SKILL.md`, the model-authoring list in `agami-connect/SKILL.md`, the declarative-fields paragraph in `docs/format-spec.md`, and the model-explorer field label and hint. The Python-side anchors stay in `#` comments, where nothing ships them.

Gate is green: 2416 passed, 4 skipped, 100% patch coverage, CI green on py3.10 / 3.11 / 3.12.

PR #199, opened 2026-08-05
ACE-083: report the aggregate a join actually multiplies, and the join that actually does it

sandeep-agami:
Principle 6c says the receipt states, for each aggregate, whether a join multiplied the rows its
value was computed from and **which join did it**. ACE-094 made that a fact rather than a refusal and
ACE-060 keyed it per aggregate. Neither changed what the analysis computes, and on `main` it gets 6c
wrong three ways. Every one is a **false receipt statement**, not a false refusal: the answer comes
back, and the sentence beside it is untrue.

**S1.** `MIN`, `MAX`, every `DISTINCT`-qualified aggregate and the boolean folds report
`risk="fan_trap"`. `runtime.py` confessed it in a comment: *"MIN, MAX and COUNT(DISTINCT) are counted
as fan-out risks although a fan-out cannot change what they return."* The rows behind that `MAX`
really **are** multiplied, so 6c's fact is true and suppressing it would make ACE-060's item claim
`not_multiplied`, falsely. What is wrong is the word *trap*, not the finding. Invariance is a second
derivable property of the aggregate node, not grounds to drop the first.

**S2.** `SUM(order_items.quantity * orders.total_amount)` reports a fan trap, because the site walk
added every table whose column appeared **anywhere** inside the aggregate. Here the value is one row
per `order_items` row and the one-side column is a scalar co-factor. Attribution has to follow the
value path, not syntactic presence.

**S3.** ACE-099 built the scope notion this needs, and `_alias_map` discarded it: it kept every
reference whatever its scope. So the analysis named a join only a CTE body takes, and missed one the
statement does take.

```
WITH oi AS (SELECT order_id, SUM(quantity) q FROM order_items GROUP BY order_id)
SELECT SUM(orders.total_amount) FROM orders JOIN oi ON oi.order_id = orders.id
  -> fan_trap naming `order_items`, a join only the CTE body takes
```

`oi` is grouped to `order_id` and joins 1:1. Nothing fans, so 6c's *"which join does it"* has no true
answer here.

## What this does

`fan_out_invariant` joins `_MULTIPLYING_RISKS`, so a duplication-invariant aggregate keeps
`status: multiplied` and nothing is labelled a trap. `AggregateReport` gains no field and
`as_dict()`'s key set is byte-identical to base, so the item shape stays ACE-060's. Attribution
follows the value path. `_alias_map` gains an **opt-in** scope filter, and a CTE reference is resolved
to what it actually stands for: a grain-preserving body binds to its source table, a grain-changing
one becomes its own entity at its `GROUP BY` grain and is compared to the join key through
`build.infer_cardinality`, and anything else is `undetermined`.

The filter is opt-in for a reason. `_alias_map` also feeds `_projected_sensitive`, a gate that
**refuses**, and `assemble_receipt`'s scope, which drives the receipt's own `joins` and `tables`.
Narrowing what a refusing gate can see is a security change this slice has no business making, so
only the fan/chasm path passes the flag and `tests/test_ace099_resolver_parity.py` stays green
unedited.

## The scope filter creates a false clean, so the binding ships in the same commit

Once `cte:*` and `subquery` references are dropped, this reads clean:

```sql
SELECT SUM(orders.total_amount) FROM orders
  JOIN (SELECT order_id FROM order_items) d ON d.order_id = orders.id
```

`d` is an `exp.Subquery` and never entered `_alias_map` at all. Today `order_items` leaks into the
outer map from *inside* the subquery, which is the only reason that shape currently reports
`multiplied`. Filter the leak and the outer scope is one table, the aggregate resolves cleanly, and a
genuinely inflated statement says `not_multiplied`. So `_reference_sites` binds FROM/JOIN-bound
non-table sources in the one existing walk, and an alias that resolves to nothing forces
`undetermined`.

## The lesson this PR is really about

Two review rounds found the same defect eight times, and every instance was one shape: **a denylist
with an unsafe default.**

The value-path walk enumerated alternation node types and let everything else union all its children.
The grain-preserving check rejected three `exp.Select` args and accepted the rest. Both defaults put
an unanticipated shape on the side that **clears a real fan**. Round one extended the enumerations;
round two found `GREATEST`, `LEAST`, `NVL2` and `DECODE` sitting just past them, with `NVL2` being
`exp.If` under another name, and `LATERAL VIEW`, `UNPIVOT` and `CONNECT BY` on the CTE side, all of
which change a row count. Every one was a regression against base: base said `multiplied`, the
half-fixed branch said `not_multiplied`.

Extending a list is not a fix when the list is the only thing holding the polarity. So both walks now
enumerate what they understand and **contribute nothing otherwise**: three node sets (alternation,
which intersects its arms, because an alternation is at a table's grain only if *every* alternative
is; structural; combining, which unions its operands), and an argument allowlist for a
grain-preserving body. The combining set was derived by instrumenting the old code and recording what
actually reached the generic branch across the suite, not from imagination. `exp.Mul` is load-bearing
in it: without it S2 stops reaching `not_multiplied`.

One measured correction to the strictest form: `ORDER BY` stays in the grain-preserving allowlist.
Excluding it loses a real finding rather than failing closed, and ordering provably cannot change a
row count.

## Availability, on caller-controlled SQL

Two holes, both found by review, both closed.

`_cte_names(tree.root())` walked the whole tree as the unconditional left operand of `&`, while
`_preflight_select` runs once per set-operation arm: O(K²). Measured at 471 arms and 49,915 characters
(just under the 50,000 cap), 0.081s became 2.021s, that one call being 93% of it. A statement with
**no CTE anywhere** regressed 37x, which made the in-code justification measurably false. Replaced by
a lexical-ancestor walk, which removes the per-arm whole-tree pass outright.

Both `_grain_preserving_source` and `_value_sources` could be driven to `RecursionError` by input
under the character cap: a 990-link CTE chain at 29,573 characters, and a wide expression at 6,496.
`execute_sql.py::_receipt_for` catches that under bare `except Exception` and returns
`RECEIPT_BUILD_FAILED`, so **the statement still executes and returns rows while the caller silently
loses the receipt**. Caller-chosen input that turns off the trust layer without turning off the
answer. The CTE side takes a depth bound; the expression walk is now iterative, following the
precedent `_and_conjuncts` already set in this file. A 12,476-term expression at exactly 50,000
characters now answers, and the test asserts that length so the premise cannot drift.

## Blast radius, stated plainly

**A shared test fixture gained declared tables.** `_sales_org` declared only `relationships`, so
`_model_table_index(org)` was empty, ACE-060's `visible` set was empty, and **no aggregate could ever
report `not_multiplied` on it**. It now declares `tables_defined` with columns and grains. Every
`Column.aggregation` stays at the default `unknown` and no metrics are declared, which is what keeps
`_check_aggregation_semantics` silent so the exact risk-list assertions do not move. Measured over all
8 consumers: 7 byte-identical, and `test_aggregating_many_side_is_allowed` moves `SUM(order_items.quantity)`
from `undetermined` to `not_multiplied`, which is the fixture becoming able to express the claim the
test's name already makes. Its assertion is strengthened to pin the status rather than rest on an
empty finding list.

**One marker clause dies, and two tests with it.** The fan-immune sentence is what S1 exists to
retire. The other four clauses stand, including the unreached-scope one, because nothing here makes an
aggregate inside a CTE reported. The two parametrize rows in
`test_each_residual_clause_is_conditional_on_the_statement` that asserted the deleted clause cannot
pass once it is gone. Both are recorded in the spec's `## Decisions`.

**Three `bad_aggregation` outcomes move**, net minus one plus two, because
`_check_aggregation_semantics` reads the one alias map this change re-scopes. Measured over a 20-shape
differential; both directions are pinned.

**Two shapes newly report `not_multiplied` and both answers are correct**: a semi-join
(`WHERE id IN (SELECT ...)`) and a scalar subquery in the SELECT list. Base was reading the subquery's
tables out of its scope. That is the S3 fix generalizing.

## Verification

3369 collected on base, 3568 now (3559 passed, 9 skipped). `dev.py check` green, `dev.py cover`
reports 100% on the changed lines.

The **reverse non-regression corpus landed first**, before any loosening, and earned it: it caught a
false clean the moment the scope filter went in without the grain resolution, before anything was
committed. It is 29 members carrying `(label, sql, inflated_aggregate_labels)` and asserts
`status != not_multiplied` over the **named** items, so a statement may hold an honestly clean
aggregate beside an inflated one. A label selecting nothing fails its own existence assertion, so a
typo cannot make a row vacuous.

The determinism test was **provably vacuous** and is not any more: deleting all three `sorted()` calls
used to leave the whole suite green. Its statement had one fanning source and one many-side table, so
no multi-element set was ever iterated. It now runs four statements, and each of the three `sorted()`
calls is caught by exactly one of them, verified by mutation.

`runtime.py` is not in `dev.py::_VENDORED` and neither `guardrail.py` nor `execute_sql.py` is touched,
so `sync-lib` is a no-op here.

## Two notes on the gate itself

`dev.py check` runs no type checker and no coverage. Coverage is enforced only in CI at
`--cov-fail-under=85`, and `dev.py cover` runs `diff-cover` against `origin/main` with no
`--fail-under`, so it effectively demands the changed lines be covered. `check` going green is not
sufficient before review.

## Known limitations, all measured and none introduced here

- A CTE that **shadows** a declared table name still over-reports: the derived `one_to_one` edge is
  appended, but the model's own `many_to_one` edge is still in the list and wins.
- `_one_side_facing_many` and `_many_side_facing_one` compare table names case-sensitively while
  `check_table_scope` folds, so an uppercase-spelled statement gets a clean receipt over a real fan.
  Same false-clean class, two functions from the case-folding fix this branch does land, and Snowflake
  and Oracle return uppercase identifiers.
- `_shared_dimension` iterates a set, so a tie between two candidate dimensions can resolve differently
  between runs. This change narrows the window but does not close it.

All three are identical on base and all three are filed as follow-ups.

## Reviewed and declined

Review argued the value-path suppression should still emit a finding naming the edge, since the rows
really were multiplied and only the value sat at the finer grain. Declined: S2 pins `joins == []` on
that shape, and a `not_multiplied` item carrying a join contradicts the item vocabulary ACE-060 owns.
The observation is fair and belongs in a spec conversation, not this diff.

Spec: ACE-083

### DIFF
--- a/packages/agami-core/src/semantic_model/runtime.py
+++ b/packages/agami-core/src/semantic_model/runtime.py
@@ -1697,7 +1697,22 @@ def _preflight_select(tree: "exp.Select", org: Datasource,
     `ctx` supplies the shared cardinality/column indices (ACE-045); `tree` is always the
     caller's own SELECT (a set-op arm ≠ `ctx.tree`), so only the indices come from `ctx`."""
     rels = ctx.cardinality_index if ctx is not None else _cardinality_index(org)
-    tables_in_scope = _alias_map(tree)  # alias -> table
+    # Filtered to what THIS SELECT's own FROM/JOIN clauses bind, and derived once: every consumer
+    # below reads this one map. Asking for it twice with two flags would be the second tree walk
+    # ACE-099 exists to prevent, and would let the fan detector and the aggregation-semantics check
+    # disagree about which tables the statement is even reading.
+    tables_in_scope = _alias_map(tree, in_scope_only=True)  # alias -> table ("" when unbindable)
+    # A name the statement bound for itself stands where a table name would, and the map cannot
+    # tell the difference: `FROM oi` reads `oi` whether `oi` is a table or a WITH. Resolving it is
+    # what lets the fan detector see the join the statement actually takes, and the guard is the
+    # cheap half of the question, so a statement with no CTE in its FROM pays nothing for this.
+    if _cte_names(tree) & {_tkey(v) for v in tables_in_scope.values()}:
+        tidx = ctx.model_table_index if ctx is not None else _model_table_index(org)
+        tables_in_scope, cte_rels = _resolve_cte_scope(tree, tables_in_scope, tidx)
+        # A NEW list. `ctx.cardinality_index` is the model's own edges, shared across every guard
+        # in the battery and across every arm of a set operation; appending a statement-derived
+        # edge to it would leak this query's CTE into the next one's analysis.
+        rels = rels + cte_rels
     table_set = set(tables_in_scope.values())
 
     sites = _aggregate_sites(tree, tables_in_scope, scope, visible)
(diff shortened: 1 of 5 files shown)


sandeep-agami:
## Rebased onto `main` @ `d24ea0f`

ACE-059 (#197) added ~659 lines to `runtime.py` while this was in review. Two hunks conflicted, both resolved so that **both specs survive** rather than by taking a side:

- **`_reference_sites`.** ACE-059 extracted the four-branch scope determination into `_scope_label`, shared with its join walk; this branch rewrote the same loop to widen it (`bind_derived`) and bind derived FROM/JOIN sources. Resolved by calling `_scope_label` from the widened loop. Taking ACE-059's side would drop `bind_derived` and re-open this branch's false clean on derived sources; taking this branch's side would leave `_scope_label` with one caller, which is the drift its own docstring says the extraction prevents.
- **`assemble_receipt`.** ACE-059 hoisted `visible` to a local; this branch added the required `tidx` keyword. Same value, two spellings, both kept.

Verified by interdiffing the pre- and post-rebase deltas: those two hunks are the only differences. Of ACE-059's 608 added lines in `runtime.py`, 606 are byte-present and the 2 missing are exactly the hand-merged ones; none of its 51 deleted lines came back.

`4213 passed, 11 skipped`, 100% patch coverage, all gates green.

## Smoke-tested end to end on real Postgres

Ran the local guardrail testbed (PG 16.12, shipped synthetic sample, real least-privilege grants) across **all three routes**: CLI, MCP stdio, and MCP http. **stdio and http were byte-identical on every vector.**

§5 least-privilege, §9 refusal contract, §9d dialect parsing and §9e unscopable sources all match the runbook exactly. That last one matters here: this branch binds the same node kinds ACE-037's `check_scopable` refuses, and all six of its vectors still refuse, so the binding did not loosen the gate.

§9c re-measured. One expectation moves and that move is this PR: `MIN(o.total_amount)` had marker *"the fan-immune clause only"* and is now **null**, because that clause is the sentence S1 exists to retire. The `aggregate in a CTE` row loses the same clause and keeps the other two. Everything else is unchanged.

New §9c-bis vectors, all confirmed on the live database: `MIN` and `COUNT(DISTINCT)` over a fan report `multiplied` with `fan_out_invariant`; `SUM(i.quantity * o.total_amount)` is `not_multiplied` with no joins; both CASE-predicate guards stay `multiplied`; `GREATEST` and `COALESCE` stay `multiplied` (the class the second review round caught); the grain-changing CTE names no join; the grain-preserving one names `order_items`; both arms of a laundered UNION report; and `MIN` / `COUNT(DISTINCT)` over a chasm still carry `chasm_trap`, not `fan_out_invariant`.

One trap worth recording for whoever runs this next: two vectors written as the spec spells them come back `refused`, by the **`SELECT *` ban**, nothing to do with this section. Project the CTE body's columns explicitly. That is the fourth time in this runbook a refusal from an unrelated rule has stood in for a gate that was not firing.

The runbook is updated with the §9c corrections and a §9c-bis block.

## Follow-ups filed, none introduced here

- #200 a CTE that shadows a declared table name still over-reports (the model's own edge outranks the derived one)
- #201 the fan/chasm helpers compare table names case-sensitively, so an uppercase statement gets a clean receipt over a real fan
- #202 `_shared_dimension` picks its dimension by iterating a set
- #203 a flaky prompt-examples test, likely `uuid4` ids with no `ORDER BY`
- #204 the `joins` and `aggregates` sections resolve a CTE reference by different rules and visibly disagree on the same edge, surfaced by this rebase

All measured identical on base except #204, which is a consequence of the two specs landing independently.

sandeep-agami:
Copilot's finding is correct and is fixed in `58e2f71`.

`_exp_nodes` exists because the package pins `sqlglot>=20` and `exp.Nvl2` / `exp.DecodeCase` are later additions. `_value_operands` then defeated it with a bare `isinstance(node, exp.DecodeCase)` in the dispatch: the attribute is read when the line runs, not at import, and the DECODE arm sits **above** the ternary and leading-argument arms. So on a sqlglot without that class, every `COALESCE`, `IF`, `GREATEST` and `LEAST` raised `AttributeError` on the way past it. `exp.Case` returns before that line, which is why the module still imported and the suite still passed.

Worth spelling out why this is more than a compatibility nit: the failure is invisible to the caller. `execute_sql._receipt_for` catches it and returns `RECEIPT_BUILD_FAILED`, so the statement executes and answers while the trust layer silently disappears. That is the same shape `_MAX_CTE_CHAIN` and the iterative `_value_sources` walk already exist for in this PR, and `COALESCE` over an aggregate is ordinary SQL.

Swept the module for the same mistake elsewhere: every other bare reference to a name-resolved class is in a comment, and `exp.Lateral` in `check_scopable` predates the pin's floor.

The regression test deletes the attribute and asserts the four alternations still attribute. Measured both ways: 4 failed against the bug, 4 pass with the fix. Gate green at 4217 passed, 100% patch coverage.

sandeep-agami:
## Smoke re-run at `58e2f71`, the tree that actually merges

No rebase needed: the branch's merge-base **is** `origin/main` (`d24ea0f`), so this is a linear extension and `58e2f71` is exactly what lands. The earlier smoke ran at `c130f73`, before the rebase over ACE-059 and before the `DecodeCase` fix, so it did not describe the current diff. Re-ran the whole thing.

**Every row is unchanged from the first pass.** That is the useful result: absorbing ACE-059's 659 lines in the same file moved nothing observable.

- **§5** least-privilege: SELECT returns 4000, all ten write/DDL/file/network paths denied by the engine with the app guard out of the loop.
- **§8** introspection: 9 relationships read-only, 12 as owner, `storage_type: PostgreSQL`. The ACE-090 defect reproduces as documented, which is what makes the chasm vector need `art_owner`.
- **§9 / §9d / §9e** on the CLI: byte-identical to the runbook. §9e is the one that matters here, since this PR binds the same node kinds ACE-037's `check_scopable` refuses, and all five vectors still refuse while all three controls still answer.
- **§9c + §9c-bis** on stdio and http: **IDENTICAL on every single vector**, 26 of them across both profiles.

New vectors this pass, all confirmed live:

- `GREATEST`, `LEAST`, `COALESCE`, `NULLIF` over the fan all report `multiplied`. These are the class the second review round caught and the class Copilot's finding would have broken on an older sqlglot. They reach three different arms of the dispatch, so all four are kept.
- The transitive grain-preserving CTE resolves two hops and names `order_items`.
- **The two-fan shape**, which a single-fan corpus structurally cannot express: `SUM(c.id * s.id)` over `customers` joined to both `orders` and `subscriptions` reports `multiplied` naming **`customers (1) <- orders (N)` only**, omitting the subscriptions edge, because `s.id` is on the value path so that edge's duplication is the value's own grain. Naming both edges was a live bug for one review round; naming neither is worse and was live for another.

The one §9c expectation that moves is still the one this PR owns: `MIN`'s marker was *"the fan-immune clause only"* and is null, because that clause is the sentence S1 retires.

Testbed updated at `649ad76` in agami-sdlc with the re-measured sha and the two new rows. Databases torn down.

PR #205, opened 2026-08-05
ACE-058: report per output column which declared metric it computes, and whether it is signed off

sandeep-agami:
`Spec: ACE-058`

## Summary

The trust receipt has five sections. Four already carry real content; **`columns` was the last one still marking itself undetermined**, which made this the whole remaining difference between the receipt ACE-088 declared and the one F11 §3 describes.

What shipped before matched metrics against the **whole statement** by substring containment (`_norm_sql`). Measured on a live Postgres 16.12 against the shipped sample before writing any code:

| Statement (against a signed-off `paid_revenue` = `SUM(total_amount) FILTER (WHERE status = 'paid')`) | Before | After |
|---|---|---|
| `SELECT SUM(total_amount) AS revenue FROM orders` | **nothing at all** about the output column | `unmatched`, marker `null` |
| the same, written with the declared binding | matched, but `{column: null, metric: …}` | `matched`, `paid_revenue`, on column `revenue` |
| the declared binding **qualified** (`SUM(o.total_amount) …`) | **no match at all** | `matched` |
| the binding inside a CTE that never reaches the output | **matched anyway** | `undetermined`, and the marker says so |
| two `UNION` arms, one governed and one hand-rolled | **one** indistinguishable entry | `main#1` matched, `main#2` unmatched |

Now every output column gets one item saying which declared metric it computes and whether that metric is signed off, or that none matched. **Nothing refuses**, nothing is rewritten, and the statement reaches the executor byte-identical.

## Changes

- **The comparison is structural, and there is one rule.** Parsed output expression equals parsed binding, both sides normalized by `_fold_unquoted_identifiers` (ACE-099's shipped normalizer) with table qualifiers stripped. The spec's lexical half was dropped: its second disjunct ("its binding against the expression") *was* the structural test, so the rule had two clauses and one decision. F11 §3 amended accordingly, recorded in §6 rather than edited silently — the check is strictly stricter, never looser.
- **One binding per metric, the one declared for this deployment's engine.** `bindings` is a per-engine dict and containment tried every value, so a Snowflake binding could be reported as the definition a Postgres statement used.
- **Keyed by `alias_or_name`, else `#<i>`** — the convention `resolve_result_units` already uses. No `_echo_expr`, and the `.sql()` allowlist stays at two entries.
- **A settled status is a claim.** `unmatched` is unreachable while a column sits behind a scope the walk does not enter (ACE-060's rule) or while any declaration went unread (ACE-059's).
- **Reads `org.cross_subject_area_metrics`**, which the previous walk could not see at all.
- **Deletes** the statement-level metric loop, `_norm_sql` and `UNDETERMINED_COLUMNS`; `_columns_marker` composes per receipt like the other four.
- **Three consumer surfaces updated** — `SERVER_INSTRUCTIONS`, the chart template and the agami-query skill each told a reader to look for "the entries whose `column` is null", which matches nothing now. Left alone they would have quietly stopped surfacing unapproved metrics rather than failing loudly.

## Found in review, fixed here

- **An unresolvable engine settled a column.** `_storage_type_of` returns None for a datasource declaring no connection or two engines, and `_metric_candidates` then returns an empty list — indistinguishable from a model declaring no metrics. Every column read `unmatched` under a null marker, asserting "every declared binding was read" on the strength of not knowing which engine's bindings to read. `_declarations_unread` now decides it once. The converse is tested too: a deployment with **no** metrics still settles, because nothing to read is not the same as reading nothing.
- **Two cost regressions.** `_own_alias_map` was resolved once per output column rather than once per SELECT (8.96 ms vs main's 4.09 ms on a 40-column statement; the tell was that a model with *no* metrics ran slower than one with 400). And matching scanned every candidate per column. sqlglot's `__hash__` derives from the same structure its `__eq__` does, so the comparison is a dict lookup. **Cost is now flat in model size: 5.39 ms at 0 metrics, 5.51 ms at 400**, against main's 4.09/4.44. Both pinned by counter-based tests; the memo test is mutation-verified.

## Checklist

- [x] `uv run dev.py check` green — **4271 passed, 11 skipped**
- [x] Every success criterion has a passing test (52 in `tests/test_ace058_metric_adherence.py`)
- [x] Every line added by this spec is covered; the 53 uncovered in `runtime.py` are pre-existing
- [x] Verified end to end on a real Postgres 16.12, **both MCP transports**, seven vectors identical on stdio and http; the runbook's own §11 vectors still diff identical with `truncated` absent
- [x] Never refuses — asserted against `guardrail.RefusalReason` being closed at three members, not by convention
- [x] Statement reaches the executor byte-identical (ACE-093); no second parser; chokepoint untouched
- [x] Decisions recorded in the spec's `## Decisions`, including the ⚠ ACE-083 coordinate
- [x] Rebased onto `main` @ `76aad28` (ACE-083), gate re-run and live vectors re-verified after


sandeep-agami:
Addressed both Copilot findings in `6592c2a`, plus a third the second one led to. All three are mutation-verified: reverting each fix makes its test fail.

**1. Schema-qualified column defeated a match** — confirmed. A column can be written `catalog.db.table.column` and sqlglot keeps each part in its own arg, so clearing `table` alone left `public.total_amount` behind and `SUM(public.orders.total_amount)` did not match a binding `SUM(total_amount)`. All three namespace parts are stripped now. Schema-qualified projections are what an introspected model's own SQL tends to look like, so this was not an exotic shape.

**2. FROM-less statements forced into `undetermined`** — confirmed, and it is the mirror of a bug already fixed on this branch: a false *unsettled* claim rather than a false settled one. `SELECT 1` reads nothing, so there is nothing it failed to read, yet it reported "could not be matched against the model's declared metrics" and put a non-null marker on a receipt with nothing left to establish. That costs the section the null marker meaning "established, here it is".

Fixed differently from the suggestion, for two reasons. The suggested discriminator was "does the SELECT have a FROM/JOIN"; `sqlglot 30` keys that argument `from_`, so `args.get("from")` returns `None` for *every* statement — a trap this module already documents at `runtime.py::_cte_source_table`, and the pin is `sqlglot>=20`, so both spellings would need handling. More substantively, FROM-presence is a proxy for the real question. Asking it directly through ACE-059's existing `_computed_relations` is strictly stronger: it also catches `SELECT SUM(total_amount) FROM orders o JOIN (SELECT 1 AS k) d ON o.id = d.k`, where the visible base table alone would have settled the column despite a derived table in scope. That case is now tested.

**3. A missing parser could poison the binding cache** — not reported, but found because finding 2's test settles a FROM-less statement: it passed alone and failed in the full suite. `_reduced_binding`'s `lru_cache` keys `(text, dialect)`, but `_HAVE_SQLGLOT` decides the answer and is not in the key, so a `None` produced while the parser was off was memoized and returned forever after — that binding permanently unreadable, and every column it could have matched permanently `undetermined`, in a process where the parser is present. Six test files patch that flag. The parser check now sits outside the memoized half.

Verification: 4276 passed, gate green, and re-verified live on Postgres 16.12 across both MCP transports — all seven receipt vectors identical on stdio and http.

PR #212, opened 2026-08-08
release: 0.6.0 — the five-section receipt, a dialect-aware guard, and mandatory audit

ashwin-agami:
Cuts **0.6.0** — 94 commits since `v0.5.3`.

## What this PR does

1. Bumps the version in the four places that must match: `.claude-plugin/marketplace.json` (×2), `plugins/agami/.claude-plugin/plugin.json`, and `packages/agami-core/pyproject.toml`. The last one is what `release-pypi.yml` publishes and what its tag-vs-version guard checks, so the release tag must be `v0.6.0`.
2. Cuts the accumulated `[Unreleased]` section into `[0.6.0] — 2026-08-08`, leaving `[Unreleased]` empty at the top.

## The changelog change is grouping only

Since v0.5.3 the `[Unreleased]` section accumulated ten `###` blocks — repeated `Added` / `Changed` / `Removed`, one set per PR that landed. They are merged into one block each, in Keep a Changelog order (Added, Changed, Removed, Fixed, Contract changes, Docs).

**No entry text changed.** Verified by sorting every non-heading, non-blank line of the section before and after and diffing: identical. The diff is heading placement and block order, nothing else.

## Why 0.6.0 and not a patch

The section carries genuine removals and contract changes, so anyone pinned `~=0.5.3` must not pick this up silently:

- `max_rows` is no longer an argument to `execute_sql`, and `--max-rows` is gone from the CLI
- `truncated` is no longer a field on a successful result
- the `model_safety` refusal rule is gone — a consumer keying on `refusal.rule == "model_safety"` stops matching
- `sm prepare` returns `{sql, findings, units}` and always exits 0; `sm preflight` returns `{findings: [...]}`
- receipt `scope` gained a set-operation arm ordinal (`main#1`), which breaks a `scope == "main"` equality branch
- `sm receipt --applied-filters` and the top-level `default_filters_applied` key are gone
- `Adapters.governance` is removed — and `Adapters` is not keyword-only, so a positional 4th argument now binds as `executor`

Pre-1.0, so a minor bump is the semver signal for those rather than a major.

## Also worth an operator's attention in this release

`AGAMI_GOVERNANCE_ENFORCED` (ACE-101) ships **off by default**, so a fresh server does not enforce table scope, column scope, the `SELECT *` ban, or the engine-mismatch check until it is turned on. The read-only guard, dangerous-function guard, statement timeout and row cap enforce in both postures. This is documented in the changelog entry, `docs/self-hosting.md` and `SECURITY.md`, and the server logs a startup warning naming the exposure.

## Verification

- `uv run dev.py check` — 4394 passed, 12 skipped; ruff check + format clean; gitleaks clean
- `python -m build --sdist --wheel packages/agami-core` — builds `agami_core-0.6.0.tar.gz` and `agami_core-0.6.0-py3-none-any.whl`
- `twine check` — both artifacts PASSED
- wheel `top_level.txt` still carries the flat-module invariant (`semantic_model`, `mcp_harness`, `execute_sql`, `agami_paths`)

## After merge

Publishing is **not** automatic on merge — `release-pypi.yml` and `release-image.yml` both trigger on `release: published`. A `v0.6.0` GitHub release still has to be created to push to PyPI and GHCR.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,7 +7,7 @@
   },
   "metadata": {
     "description": "The trust layer between AI and your data. A governed semantic model for AI-to-database access — every join approved, every metric signed off, every answer with a receipt. Local-first: no infra, no servers, no data leaves your machine.",
-    "version": "0.5.3",
+    "version": "0.6.0",
     "pluginRoot": "./plugins",
     "tags": ["data", "sql", "governance", "trust-layer", "semantic-model", "local-first", "fair-code"],
     "repository": "https://github.com/AgamiAI/agami-core",
@@ -18,7 +18,7 @@
       "name": "agami-core",
       "source": "./plugins/agami",
       "description": "A governed trust layer between AI and your database: every join FK-derived or human-approved, every metric signed off, every answer with a receipt (the SQL, the model version, who vouched for each definition). Runs entirely on your machine via Claude's Bash / Read / Write tools — no MCP server or Python required if you have psql/mysql or DuckDB (an optional local MCP server is available for use from Claude Desktop).",
-      "version": "0.5.3",
+      "version": "0.6.0",
       "keywords": ["database", "governance", "trust-layer", "semantic-model", "sql", "postgres", "snowflake"],
       "category": "data"
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ below corresponds to one such version.
 
 ## [Unreleased]
 
+## [0.6.0] — 2026-08-08
+
 ### Added
 
 - **A server can now run with the semantic-model pass off, and always says so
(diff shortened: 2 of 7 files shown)


ashwin-agami:
## Update: the review found a release-blocking upgrade defect, now fixed in this PR

`/agami-sdlc:review` (low lane → `/code-review`) surfaced three findings. Two are addressed here; the third is a judgement call left for the code owner.

### 1. Fixed — updating the plugin did not update the library it runs on

`ensure_pkg` in `plugins/agami/scripts/sm` returned early on `_imports_ok`, which only asks whether the library *imports*. A stale one does. And `_agami_lib.ensure_importable()` explicitly prefers the pip-installed package over the fresh copy bundled in `lib/`, so the plugin shipping a correct 0.6.0 library does not help.

Concretely, before this fix: a user with `agami-core` 0.5.3 installed who updates to plugin 0.6.0 gets 0.6.0 skills against the 0.5.3 library. v0.5.3's `assemble_receipt` emits `tables_used / relationships / metrics / named_filters / assumptions / warnings`; 0.6.0's `render_chart.py` `_validate_receipt` requires `columns, tables, joins, aggregates, assumptions` and raises. Every charted query fails with `receipt.columns is missing`, silently until first run, with nothing naming the cause.

This is a pre-existing defect in the installer — the line is not otherwise in this diff — but 0.6.0 is the first release that detonates it, so it is fixed here rather than filed.

Readiness is now `_imports_ok && _version_ok`. The floor is `$VER`, the plugin's own version, because the plugin and the package are bumped to the same string on every release — no second constant for a release to forget. A stale install is reinstalled with `--upgrade` (pip counts a present distribution as satisfying a bare requirement, so it would otherwise no-op), and the version is re-checked after each source, so an index that has not published this release yet falls through to git rather than passing as done. A library at or above the floor is left alone; a source layout with no distribution metadata is treated as satisfied rather than churned.

Tests: `test_stale_library_is_upgraded_not_left_alone` and `test_current_library_is_left_alone`. Mutation-checked — the stale test fails against the old script (it attempted no install at all), so it catches the actual bug rather than restating the new code.

### 2. Fixed — the release checklist omitted the file that publishes

`CONTRIBUTING.md` still said "three places across two files" and did not list `packages/agami-core/pyproject.toml`, which has been the fourth place since 0.5.3 and is what `release-pypi.yml` publishes and checks the tag against. Now listed, plus a note that merging a bump publishes nothing — both release workflows fire on `release: published`.

### 3. Not actioned — semver, for the code owner to settle

`CONTRIBUTING.md:111` says removed flags / anything breaking an install → **major**, and line 113 says not to fold a breaking change into a minor. 0.6.0 does exactly that. The pre-1.0 reading (no stability promise below 1.0, which is how 0.5.0 shipped) says a minor is fine; the table as written says otherwise. `~=0.5.3` pins are safe either way, `<1.0` bounds are not. Left as 0.6.0 deliberately — if the rule is meant literally, the fix is to reconcile the table rather than to re-cut this release.

### Gate

`uv run dev.py check` — **4397 passed**, 12 skipped (up 3, the new tests); ruff check + format clean; gitleaks clean.

PR #214, opened 2026-08-08
release: 0.6.1 — a refusal's sentence reaches the audit row again

ashwin-agami:
Cuts **0.6.1** — one commit since `v0.6.0` ([#213](https://github.com/AgamiAI/agami-core/pull/213)).

## What this PR does

1. Bumps the version in the four places that must match: `.claude-plugin/marketplace.json` (×2), `plugins/agami/.claude-plugin/plugin.json`, and `packages/agami-core/pyproject.toml`. The release tag must be `v0.6.1` or `release-pypi.yml`'s guard fails the publish.
2. Adds the `[0.6.1]` changelog section for #213, which shipped without one.

## Why a patch bump

#213 changes *what gets recorded*, not any interface: no argument, return shape, refusal rule, receipt key or CLI flag moves. `tool_calls.refusal_detail` / `refusal_remediation` go from NULL to populated on a real refusal. Nothing a consumer branches on changes meaning — a row that was empty starts carrying the sentence it was always supposed to.

## Changelog accuracy

I verified the entry against the code rather than the PR summary:

- **"the two columns added in 0.6.0"** — `018_tool_calls_refusal_sentence.sql` is present at the `v0.6.0` tag. Confirmed.
- **"kept when the stated outcome agrees with the body"** — [tools.py:2125](https://github.com/AgamiAI/agami-core/blob/main/packages/agami-core/src/tools.py#L2125) clears on `derived_success or derived_error_kind != body_refusal_rule`, so they survive only when the call is unsuccessful *and* the stated kind equals the rule the body named. Matches.
- **"four tests, each mutation-checked"** — this phrasing from #213's summary is **not** in the entry. Four tests at `v0.6.0` touch refusals, but only two (`test_a_refusal_records_the_gate_s_own_sentences`, `test_a_long_refusal_detail_is_bounded_before_it_is_stored`) cover these sentences, and mutation-checking isn't a property I can verify from the tree. The entry states what is true of all of them and checkable: each drove the recorder on a path no production caller takes.

## Verification

- `uv run dev.py check` — 4399 passed, 12 skipped; ruff check + format clean; gitleaks clean
- `python -m build --sdist --wheel packages/agami-core` — builds `agami_core-0.6.1.tar.gz` + `.whl`
- `twine check` — both artifacts PASSED
- `0.6.1` confirmed free on PyPI

## After merge

Merging publishes nothing. Both release workflows fire on `release: published`, so a `v0.6.1` GitHub release still has to be cut to reach PyPI and GHCR.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -7,7 +7,7 @@
   },
   "metadata": {
     "description": "The trust layer between AI and your data. A governed semantic model for AI-to-database access — every join approved, every metric signed off, every answer with a receipt. Local-first: no infra, no servers, no data leaves your machine.",
-    "version": "0.6.0",
+    "version": "0.6.1",
     "pluginRoot": "./plugins",
     "tags": ["data", "sql", "governance", "trust-layer", "semantic-model", "local-first", "fair-code"],
     "repository": "https://github.com/AgamiAI/agami-core",
@@ -18,7 +18,7 @@
       "name": "agami-core",
       "source": "./plugins/agami",
       "description": "A governed trust layer between AI and your database: every join FK-derived or human-approved, every metric signed off, every answer with a receipt (the SQL, the model version, who vouched for each definition). Runs entirely on your machine via Claude's Bash / Read / Write tools — no MCP server or Python required if you have psql/mysql or DuckDB (an optional local MCP server is available for use from Claude Desktop).",
-      "version": "0.6.0",
+      "version": "0.6.1",
       "keywords": ["database", "governance", "trust-layer", "semantic-model", "sql", "postgres", "snowflake"],
       "category": "data"
     }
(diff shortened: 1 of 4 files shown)


ashwin-agami:
Addressed the inline comment in `4a0b6f1`'s successor — both columns now carry the `tool_calls.` qualifier.

Also corrected two claims of my own that `/code-review` caught in the changelog prose, which were the same class of overreach this PR's description flags in #213's summary:

- *"Every test covering these sentences ... stated nothing"* was **false**. `test_an_overridden_outcome_drops_sentences_parsed_from_the_body` at `v0.6.0` states `error_kind="stated_by_the_caller"`, and it is the test that encoded the buggy unconditional clear. The entry now claims only what holds: every test asserting the sentences were *present* stated nothing.
- *"The new tests ... fail against the shipped behaviour"* holds for **one** of the two. `test_a_stated_success_drops_them_however_the_body_read` asserts NULL, which 0.6.0's unconditional clear already produced, so it passes pre-fix. Only the agreement test fails against shipped behaviour.

PR #215, opened 2026-08-09
A CASE predicate is not what a SUM aggregates, and a shared expression names no metric

sandeep-agami:
## Summary

Two defects on the trust receipt. Each made it state something false about an answer, and together they made one response **contradict itself**: the `aggregates` section called an expression meaningless while the `columns` section of the same receipt matched that identical expression to an approved, signed-off metric. A reader cannot act on both.

## 1. The aggregate's predicate was read as its argument

`_bare_aggregate_column` promised the column an aggregate is applied to *"ONLY when the argument is that bare column"* and then walked the whole subtree with `find_all(exp.Column)`, descending into `CASE` branches. Its `exp.Binary` guard caught `=` and `IS NULL` by accident and missed `IN` and `BETWEEN`, which are not `Binary` subclasses.

So `SUM(CASE WHEN state NOT IN (…) THEN 1 ELSE 0 END)` yielded `state` and the receipt reported *"SUM(state) is meaningless"* about an expression the statement never wrote — while the same conditional count spelled with `=` passed. The rule's intent is right; the extraction was wrong.

It now reads the aggregate's **direct argument**, unwrapping a single-expression `DISTINCT` (`COUNT(DISTINCT a, b)` aggregates a tuple, not a column).

## 2. A binding two metrics share identified neither

`_by_binding` kept only the first candidate per reduced binding (`setdefault`) and called that REQ-022 satisfied. It is not. REQ-022 says two names for one expression is a model-authoring question and *not one this layer decides* — and silently picking the first **is** deciding it.

A model that auto-suggests one `*_count` per table declares `COUNT(*)` once per table, so on a wide model `COUNT(*)` resolved to whichever metric happened to sort first, and the receipt reported one table's count as an unrelated table's count metric: `matched`, `confirmed`, `approved`. One authoritative false attribution is worse than dozens of obvious ones — it is indistinguishable from a true one, on the surface whose whole job is to say which declared metric an answer computes.

The index now carries **every** candidate. When several share a binding, the tables the statement reads discriminate; when they cannot narrow it to exactly one, the answer is `undetermined`. The guard **never rejects** — a candidate declaring no `source_tables` is not eliminated, because an absent declaration is not evidence against a metric. `_strip_qualifiers` named this guard as the thing to add *"if a false match shows up"*; one has, and unlike when that was written the colliding metrics do carry `source_tables`.

The `aggregates` section already reported `undetermined` for the same `COUNT(*)` (`_AggSite.resolved` is false with no columns). One section of the receipt was already right; this makes the other agree.

## A deliberate spec change

`test_two_metrics_declaring_one_binding_resolve_the_same_way_every_run` pinned first-wins as a determinism guarantee. Two metrics sharing a binding with nothing to tell them apart now read `undetermined` — still deterministic, and closer to REQ-022's own words. The test is rewritten to state the new contract. `test_matching_does_not_scan_every_declared_metric` changes mechanically (the map's value is a list); its intent — one hash per column, not a scan — is preserved.

## Verified against a real model, before and after

Not only fixtures. The same 277-metric model and the same two statements, run against `main` and against this branch:

```
                                    main                          this branch
SUM(CASE WHEN state NOT IN (…))  bad_aggregation:              (no findings)
                                 "SUM(state) is meaningless"
  ...its columns entry           matched -> the approved       unchanged
                                 metric, signed off            -> the two sections agree

COUNT(*) AS <alias>              matched -> an unrelated       undetermined, no metric
                                 table's count metric          named, marker explains why
```

## Checklist

- [x] `pytest` — 4411 passed, 12 skipped
- [x] `ruff check` + `format --check` clean; `dev.py check` green
- [x] New tests **fail against the shipped behaviour** (five CASE spellings, incl. the two that passed only by accident)
- [x] Every changed function fully covered — lines **and** branches, verified by intersecting the miss list against each function's AST range
- [x] `runtime.py` is not vendored, so `test_vendored_surface_parity` is unaffected (it compares `(status, rule, reason)`, never findings)
- [x] Verified against a real model, not only the suite
- [x] No real customer data, names or credentials

### DIFF
--- a/packages/agami-core/src/semantic_model/runtime.py
+++ b/packages/agami-core/src/semantic_model/runtime.py
@@ -2319,13 +2319,24 @@ def _lookup_column(col: "exp.Column", scope: dict[str, str],
 
 
 def _bare_aggregate_column(agg: "exp.AggFunc") -> Optional["exp.Column"]:
-    """The single column an aggregate is applied to, ONLY when the argument is that
-    bare column (optionally DISTINCT). Returns None for composite args like
-    SUM(price * qty) — those can be legitimately additive even if a part isn't."""
-    cols = list(agg.find_all(exp.Column))
-    if len(cols) == 1 and agg.find(exp.Binary) is None:
-        return cols[0]
-    return None
+    """The single column an aggregate is applied to, ONLY when the argument IS that bare
+    column (optionally DISTINCT). Returns None for composite args like SUM(price * qty) —
+    those can be legitimately additive even if a part isn't.
+
+    Reads the aggregate's DIRECT argument. It used to walk the whole subtree
+    (`find_all(exp.Column)`, guarded by `find(exp.Binary) is None`), which reads a column out
+    of a CASE *predicate* and reports it as the summed value: `SUM(CASE WHEN state NOT IN
+    (6,7,8) THEN 1 ELSE 0 END)` returned `state`, and the caller published "SUM(state) is
+    meaningless" about an expression the statement never wrote. The `exp.Binary` guard caught
+    `=` and `IS NULL` by accident and missed `IN`/`BETWEEN`, which are not Binary subclasses —
+    so the same logical query was flagged or cleared on which operator the CASE happened to
+    use. A subtree walk cannot answer "what is being aggregated"; only the argument can."""
+    arg = agg.this
+    if isinstance(arg, exp.Distinct):
+        # DISTINCT holds a LIST — `COUNT(DISTINCT a, b)` aggregates a tuple, not a column.
+        exprs = arg.expressions
+        arg = exprs[0] if len(exprs) == 1 else None
+    return arg if isinstance(arg, exp.Column) else None
 
 
 def _semi_additive_columns(org: Datasource) -> dict[tuple[str, str], "Metric"]:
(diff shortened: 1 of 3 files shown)


sandeep-agami:
### 6 — the docstrings put words into REQ-022's mouth ✅ fixed in `d4a2f0a`

Correct, and I checked it rather than taking it on trust. Every REQ-022 citation in the file is receipt determinism — `runtime.py:1246` ("the same receipt on every run for the same statement"), `:2858`, `:3073`, `:4152` — and in the original docstring the `(REQ-022)` marker attached to exactly that sentence:

> First-wins in candidate order, so two metrics declaring the SAME binding always resolve to the same one and the receipt is identical on every run **(REQ-022)**. Two names for one expression is a model-authoring question and not one this layer decides; …

The second sentence was the prior author's own uncited rationale, and my rewrite promoted it into something REQ-022 states. It doesn't — and both spellings satisfy REQ-022 anyway, since a stable first-wins pick and a stable `undetermined` are each identical on every run.

The docstring and the test now say the true thing: first-wins answered a question this layer cannot answer — *which* of several metrics declaring one expression a column computes — and declining to answer it is a deliberate contract change, which is how the PR body already framed it. No behaviour change; `4411 passed`.

sandeep-agami:
@copilot correct, and worse than described — fixed in `bd4e7f2`.

`_tkey` only lowercases; it does not strip a schema. The written side of the comparison is bare table names off `_own_alias_map`, so `public.orders` never intersected a statement reading `orders`. Confirmed:

```
bare_name('incident.incident') = 'incident'
_tkey('incident.incident')     = 'incident.incident'
```

**The consequence isn't limited to a false `undetermined`.** A candidate declaring no `source_tables` is deliberately never eliminated — so when the correctly-declared one is wrongly dropped, the undeclared one is left standing *alone* and gets **named**, `matched`, with whatever sign-off it carries. That is the exact false attribution this guard was added to prevent, reintroduced through a mismatch between the two halves of one comparison. Reproduced before fixing:

```
E  AssertionError: the qualified declaration was dropped and the undeclared metric was named in its place
E  assert [('revenue', 'matched', 'unscoped_twin')] != [('revenue', 'matched', 'unscoped_twin')]
```

Both sides now normalize with `bare_name`, which is the rule `loader._metrics_for` has always applied to this same field. Two tests, both failing before the change — the false-`undetermined` half and the false-match half, kept separate because only one of them is dangerous.

Worth recording why this survived every check I ran: **no model I have qualifies `source_tables`** — 0 of 277 metrics in the model I probed against — so the unit tests, the coverage sweep and four live deployments all passed over it. A static read caught what execution could not.

PR #216, opened 2026-08-09
The tool surface reports what it checked, and accepts what it advertises

sandeep-agami:
## Summary

Nine findings from a connector audit, all one shape: **the server told the agent something it had not established, or named a field the agent could not reach.** None of them is exotic; each shipped because nothing compared the two halves.

## Stated without checking

- **`model_present` was the literal `True`** on the served path (`tools.py`), where the listing is built *from* the model rows and it could never be false. A constant presented as a check implies a verification that did not happen. Dropped there. It survives locally, where it really does answer *"has this profile been introspected yet?"* — and that asymmetry now has a test on **both** sides.
- **`resolve_profile` could never name a real profile on a served deploy.** It resolved explicit → `AGAMI_PROFILE` → `.config.active_profile` → the literal `'default'`, and a DB-only deploy never populates `.config`. Three things followed from that one string: the listing reported `active_datasource: "default"` beside the real datasource; `is_active` was *structurally* incapable of being true; and omitting `datasource` — which every tool description says defaults to the active profile — refused with `model_unavailable`. It now consults the store, and only when **exactly one** datasource is served: with two there is no defensible answer and guessing is worse than the refusal.
- **A hosted server claimed "All execution is local."** That is a privacy claim, and on a hosted deployment it is false — the SQL runs in the container and the rows come back over the wire. It shipped verbatim on both because a module constant evaluated at import cannot see the environment that decides which is true. `server_instructions()` branches that one sentence on `_hosted()`; everything after it is shared, so the flow cannot drift, and `extra_instructions` still appends rather than replaces so the PII rule cannot be composed away.

## Promised but not sent, or accepted but not advertised

- **`bindings`** — the server instructions *and* this tool's own description both say to reuse a metric's `calculation`/`bindings` **VERBATIM**, and the metric block did not contain them. The agent was told in capitals to reuse a field it never received, so it hand-rolled from the prose `calculation` instead — which also costs the receipt a true match, since hand-rolled SQL does not reduce to the declared binding. Now sends the binding for **this deployment's engine**, not the whole per-dialect dict.
- **`area` on `get_prompt_examples`** — the handler passed it and `select_examples` implemented it correctly (`area = ? OR area IS NULL`, the area *plus* the cross-area bucket). It was missing from the `inputSchema`, which sets `additionalProperties: false`, so no compliant client could send it: a finished, correct feature, dead on every call. Making it reachable exposed that the **local** path globbed every area and ignored it, so it narrows there too — one schema, one behaviour.
- **`description` on `list_datasources`** — the tool exists to answer *"which datasource does this question touch?"*, and returned nothing to answer it with, so an agent had to pull a full schema per candidate purely to choose. The column already existed on `datasource_model` and was already written; this is one grouped read, mirroring `model_table_counts`.
- **`for_questions_about` has no writer anywhere** — `setdefault`-ed empty and never filled, including in the sample model this product ships. Dropped from the payload; cross-area edges now name their **endpoint tables**, which is what distinguishes edges between the same pair of areas. The join mechanics stay off this tier on purpose: a `dataset_names` call already returns each edge in full via `_relationships_among`.

  The field stays **declared** on the model. These models `forbid` unknown keys and generation wrote `for_questions_about: []` to disk, so deleting it would fail every such model at load. Removing it needs a format migration, not an edit.

## The contract tests were the real defect

`test_contracts.py` validated hand-written samples — written to agree with the contract, so the pair could never disagree and never caught anything:

| contract | declared | actually sent |
|---|---|---|
| `SubjectAreaSummary.tables` | `list[str]` | `[{name, description}]` |
| `CrossAreaRelationship.for_questions_about` | `str \| None` | `list[str]` |
| `DatasourceSchemaResult.large_tables` | *(absent)* | `{table: row_count}` |

The third was found by the **new** test on its first run, which is the whole argument for it. The contract is now validated against what the tool actually emits, in each mode.

`test_tools_registry.py` gains the general form: every `args.get("x")` a handler reads must be declared on that tool's own schema. That is what makes the `area` class of defect fail in CI instead of shipping unreachable.

## Docs

`docs/format-spec.md` listed `default_time_window` among the fields *"applied generically by the MCP/runtime"*. Nothing reads it — and unlike `default_filters` and `sensitive`, it is not reported in the receipt either, so it needed a category the spec did not have rather than a move into an existing one.

## Checklist

- [x] `pytest` — 4435 passed, 12 skipped, coverage 92%
- [x] `ruff check` + `format --check` clean; `dev.py check` green
- [x] New tests fail against the shipped behaviour
- [x] Every changed function fully covered — lines **and** branches, verified by intersecting the miss list against each function's AST range
- [x] Both halves of every deliberate asymmetry tested (served vs local `model_present`; hosted vs local instructions; `area` on both paths)
- [x] Absent-vs-null pinned where a field is optional (`binding`, `description`)
- [x] No model already on disk fails to load — the deprecated field stays declared
- [x] No real customer data, names or credentials


sandeep-agami:
Checked all six against the code. **Five are correct and are fixed in `5c5e731` / `d4a2f0a`; #7 I'd push back on.**

### 2 — `_sole_served_datasource` pins its own failure ✅ fixed

Correct, and the worst of the set: the failure is silent and permanent. Now caches **only a success** in a plain dict, so a negative stays retryable. Two tests: a store that throws on the first call and recovers must be seen on the next, and a success must still be memoized (a Store connection per tool call is what the cache is for). `cache_clear` is kept as an attribute so tests reach for it the same way they do on `resolved_org_id`.

### 3 — the `setdefault` removal is a no-op ✅ fixed, and you're right about why

Confirmed by running it: `for_questions_about` defaults to `[]`, not `None`, so `exclude_none` never dropped it and the dump always carried the key. The comment claimed something the code did not do.

Taken the first of your two options rather than the second — `_model_dump` now excludes it by name, so the claim becomes true instead of being walked back. Pydantic ignores an `exclude` key a model lacks, so the one call site covers every model type there. The field stays declared, so anything already on disk keeps loading.

### 4 — `model_present: bool = False` ✅ fixed

Right, and it is the same defect one layer down — a served entry validating to "no model" for a datasource that definitionally has one. Now `bool | None = None`.

### 5 — the seam is verified only on the local path ✅ fixed

Correct, and the wrong half: `extra_instructions` is a hosted-deployment seam. Both `test_tool_extension_seam` assertions and the three ACE-042 surface checks now run against `server_instructions()` in **both** modes (parametrized / via a small `_instruction_variants()` helper), including the drop-directives test — the hosted preamble is new text and has to satisfy the no-spec-ids and filter-warning rules too.

### 7 — nine defects in one PR ⚠️ pushing back, with a caveat

They aren't as separable as they look. Three of them (`model_present`, `description`, the active-profile resolution) are the same dict literal in `tool_list_datasources`; `for_questions_about` and the contract-type fixes are the same `CrossAreaRelationship` declaration; and the contract test that catches drift only exists because those types changed. Split four ways, three PRs would be stacked on unmerged bases and touching the same two files.

Where I think you're right: **`bindings`** (`_metric_full`) is genuinely independent of the rest and could stand alone. Happy to lift that one out if you'd like it reviewable on its own — say the word and I'll re-cut.

sandeep-agami:
@copilot correct — fixed in `a91d6c8`.

`(x or "").strip()` raises on any **truthy** non-string; `0`, `None` and `""` fall through the `or` safely, `True` and `1` do not:

```
area=True: AttributeError — 'bool' object has no attribute 'strip'
area=1:    AttributeError — 'int' object has no attribute 'strip'
```

And your reachability point is the reason it matters: the handler is called directly by tests and embedders, not only through a schema-validating transport.

It is also specific to this change rather than pre-existing — before `area` was honoured on the local path the argument was ignored entirely, so no input could crash it. Honouring it made a total function partial.

Guarded with `isinstance`, and a non-string is treated as **no scope** rather than refused: this path returns the curated library and has no vocabulary for an input error, so the caller gets what they got before. Parametrized over `True, 1, 0, [], {}, 3.5`, plus a whitespace-only case beside it — `"   "` must read as "no scope given", not "an area named nothing", which would match no directory and return the empty note.

PR #217, opened 2026-08-09
get_datasource_schema answers within the scope the caller declared

sandeep-agami:
## Summary

Two defects, one cause: the tool did not treat a declared scope as meaning anything.

**It discarded what a scoped call had already resolved.** `_table_contexts` asked `get_table_context` for relationships and metrics, then kept only `ctx["tables"]`. So the call whose entire purpose is per-table detail returned columns and **no way to join them** — while this tool's own description promised *"…default_filters, **relationships**, caveats, value_transforms, **metrics**"*. Same defect class as the `bindings` gap in #216: a field the surface names and never sends. In practice an agent could only write the join if a human had put it in the prose narrative.

**And it sent the whole metric catalogue on every call**, under a never-hide guarantee introduced by ACE-001 for the *verbosity ladder* — where the client asked for everything and silently got less. A scoped call is the opposite situation: the caller said which tables they're working on, so a metric from an unrelated area prevents no divergence there.

## The rule

**Within the scope you declared, nothing is hidden** — the scope's own metrics, plus the cross-area bucket (which belongs to no area, so it is out of scope nowhere; `get_prompt_examples` already scopes this way with `area = ? OR area IS NULL`).

**Scope is declared, never inferred.** `area` (new — completing the set `execute_sql` and `get_prompt_examples` already had) and `dataset_names` narrow. `query` ranks and `metric_names` selects detail; narrowing on either would silently stop showing a caller the rest of their datasource when they never said they were only working in part of it. The response echoes the `scope` it resolved, as a **declared** contract field — `extra="allow"` means an undeclared echo would round-trip silently and prove nothing.

## Measured on a 52-table, 277-metric model

| call | before | after | metric_index | joins | metric defs |
|---|---|---|---|---|---|
| no args | 60,049 B | 57,353 B | 277 | — | — |
| `area="incident"` | *parameter did not exist* | 25,483 B | **8** | — | 8 |
| `dataset_names=["incident"]` | 39,559 B | **28,035 B** | **7** (was 277) | **5** (was 0) | **7** (was 0) |
| `query=…` | 60,049 B | 57,929 B | **277 — unchanged** | — | 1 |
| `metric_names=…` | 60,049 B | 57,929 B | **277 — unchanged** | — | 1 |

**One number that goes the other way, stated plainly:** `area="request"` returns 63,543 B — *more* than an unscoped call. One area is ≤12, so it sizes to `full`, which adds the per-table column block an unscoped `summary` omits. A table-rich area trades breadth for depth. The JSON stays under the 60K budget (39,802 B) so the ladder is satisfied, and it is the content the caller asked for — but "everything got smaller" would be false.

## Review panel found six real defects — five of them hides

`/review` (correctness + structural rubric) was run before this PR, and each finding below shipped with a repro. They are in the second commit, separately reviewable:

- **`_areas_owning` read only `tables_defined`** — a table DEFINED in one area but REFERENCED by another through a `TableRef` (a shared dimension) put the referencing area's un-sourced metric out of scope. By this change's own rule such a metric "might apply to any table in its area". No fixture in the suite had a shared table.
- **`_table_contexts(area=…)` was itself a contradiction** — two reviewers proposed opposite fixes because of it. It forced every named table into the declared area (so `area` + `dataset_names` returned "not found in scope" for a table outside it) while `_scoped_metrics` ignored `scope.area` entirely: columns vanishing for a table whose metrics stayed advertised. The per-table lookup was already correct; the parameter is deleted.
- **An unknown `area` returned an empty model** under a scope echo claiming nothing was hidden — vacuously true, and indistinguishable to an agent from "this datasource has no areas". Now a `not_found` naming the areas that exist.
- **`mode=auto` was sized by the whole datasource** — an area-scoped call on a 60-area model started at `index`, which lists no tables, narrowing content while defeating the point. The ladder and budget are untouched (both out of spec scope); only the count fed to the selector is scope-aware.
- **A cross-area edge spanning two scoped areas was emitted twice** — tables resolve one group per area and `_relationships_among` returns the edge for either endpoint.
- **`with_area_list=False` was unconditional** but its justification is "the JSON already carries it", and the table tier emits no `subject_areas` — so the listing was dropped from both surfaces at once. Now conditioned on the JSON actually carrying it.

Simplicity budget: `_bare_name` duplicated `models.bare_name` (whose own docstring calls itself the single source of truth) and `_resolve_scope` inlined a third copy of the same fold. The stated reason — avoiding a module-scope model import — doesn't survive the file, which makes lazy `semantic_model` imports at eight sites. Now delegates.

## A correction to #216

A comment I merged in #216 claimed a `dataset_names` call *"already returns the join columns, `on`, cardinality and trust block in full, via `_table_contexts`"*. **That was false** — it is the bug this PR fixes. The comment now says so. The org-level edge list's design still stands on its other legs.

## Verified end to end against the deployed image

Deployed to the testing service (rev `00010-mf4`) and driven with five NL questions through the full flow — orientation → area scope → table scope. **70 assertions, 0 failures.** The acceptance criterion is the last one: SQL built *purely* from the response — join keys from `relationships`, expression from `metrics[].binding`, no prose consulted — and the receipt reads it back as `matched` to the declared metric with a `declared` join and no false aggregate finding.

## Checklist

- [x] `python3 dev.py check` — 4488 passed, 12 skipped; ruff, format, gitleaks clean
- [x] New tests fail against the shipped behaviour; two synthetic fixtures cover shapes no real model contains (a metric with no `source_tables`; a cross-area metric naming a scoped table)
- [x] Every review finding lands with a regression test, including the `TableRef` shared-table hide nothing previously exercised
- [x] One of my own tests was vacuous (compared two response lengths, true before the change) and now asserts the catalogue is a proper subset
- [x] No test weakened or deleted — `test_get_datasource_schema_sizing.py` is additive; `test_query_ranks_and_limits_metrics` still pins the full 4-metric index as the standing "query does not scope" proof
- [x] Surface-neutral: one shared `TOOLS` registry, both transports
- [x] No real customer data, names or credentials — fixtures are `acme`/`orders`/`users`/`sales`

Spec: ACE-107


sandeep-agami:
## Parameter interaction — two more defects, same class

Probed all eleven combinations of the four parameters. Two defects, both "declared but inert":

**`area` was inert at table scope, and the scope echo reported it anyway.**

```
area="change" + dataset_names=["incident"]
  -> scope: {"level":"table", "area":"change", ...}   ← false
  -> metric_index: incident's 7 metrics                ← not change's
```

`incident` isn't even in the `change` area. The echo exists so the guarantee's boundary is legible; an echo that can lie makes the guarantee unverifiable.

`area` now **validates** at table scope — the two arguments are one compound declaration ("these tables, in that area"), so a table outside the area makes them contradict and is a caller error naming both halves. It deliberately does **not override** the per-table lookup: that was the earlier revision the review panel caught, which returned `"not found in scope"` for the table while its metrics stayed advertised.

**`query` / `metric_names` were silently dropped at table scope.** They narrowed the detail block at datasource and area scope and did nothing one tier down — same argument, two behaviours, no signal. Resolved once now for every tier. Neither narrows the *scope*: `metric_index` still lists everything in it, and a selector naming an out-of-scope metric cannot drag it in (tested).

### After

| combination | scope | echo area | metric_index | detail |
|---|---|---|---|---|
| `area` + `dataset_names` (agree) | table | `sales` | 7 | 7 |
| `area` + `dataset_names` (**contradict**) | — | — | **`not_found`, names both halves** | — |
| `dataset_names` + `query` | table | — | 7 | **1** (was 7) |
| `area` + `query` | area | `sales` | 8 | 1 |
| all four | table | `sales` | 7 | 1 |
| `query` alone | datasource | — | 277 | 1 |

Six new tests, one per combination class. `dev.py check`: **4494 passed**, ruff/format/gitleaks clean.

Found by review question rather than by a failing test — worth noting, since the probe battery in the PR body exercised each parameter alone and all of it passed.

PR #223, opened 2026-08-11
feat(mcp): the surface states what the server already does

sandeep-agami:
Four capabilities shipped working and undocumented. A broken feature gets reported; a silent one just never gets called.

Came out of a side-by-side audit of the MCP instruction and tool surfaces across `agami-core`, `agami` and `agami-data-agent`. Most of what looked like a capability gap against data-agent turned out to be a description gap here.

### `execute_sql` documents its third status (Tier 1 #1)

The description named `ok` and `refused`. The code also returns `status:'failed'` carrying a `Failure{kind, message}`, with `kind` from the ten-value `guardrail.FailureKind`. A client was never told that channel exists, so an agent met an unfamiliar payload shape at the one moment it could least reason about it.

Now documented, with the split that decides whether to retry at all:

- `syntax` / `column_not_found` / `table_not_found` are repairable from the schema already held. Fix and retry silently, without narrating or echoing the failed SQL.
- `auth` / `dsn` / `driver_missing` / `permission` are deployment configuration. Retrying cannot help, so say what is wrong and stop.

Four of the ten kinds remain declared-but-unreachable (they need driver-text parsing). They are named anyway: a closed set a client can branch on exhaustively is the point, and a set that grows silently under the reader is not closed.

### `list_datasources` describes the fields it returns (Tier 1 #2, #4)

`database_type` and `table_count` have always shipped on every entry and were described nowhere, so the dialect was guessed on every non-metric expression and the schema call was made blind to how big the answer would be.

The shared instructions now say to take the dialect from `database_type` and never assume one. A metric's `binding` still arrives pre-resolved to that dialect, so it is copied not translated; the rest of the statement never was, and date arithmetic, string functions and casts are where a guess shows up.

`table_count` lets an agent predict which verbosity `get_datasource_schema` will land in before calling it, so it can scope with `area` up front instead of paying for a downgraded payload and re-asking.

### `get_datasource_schema` names its table context (Tier 1 #3)

`_table_contexts` requests `default_filters`, `relationships`, `caveats` and `value_transforms` on every call and the description named none of them. So a declared filter was first *met* on the receipt, after the statement it belonged in had already run. It is now named where it is readable while the SQL is being written.

### Two rules the instructions never carried (Tier 1 #7, #9)

- The two grounding calls share no state. Stated as a prohibition rather than a permission, because an affordance can be declined and a rule gets checked against: *never serialize what is independent*.
- There is no tool here for saving a correction, deliberately, since that is a skill operation. Absence and omission are indistinguishable to a reader, so the surface now says so and tells the model to stop claiming it remembered something it had no way to save.

### The receipt moves to the tool that returns it (Tier 1 #5)

The field-by-field vocabulary now lives on `execute_sql`'s description, where a host renders it beside the result it describes. Read at initialize, four steps before the payload exists, it was the least-situated text on the surface.

What stays in the instructions is what outlives any field name: show the unapproved joins and metrics, answer-and-warn rather than refuse, and never read an empty `items` with a marker set as clean.

**Honest accounting:** the instruction block drops 5,817 to 4,078 chars, but total always-on context grows (8,362 to 11,804), because tool descriptions are also always-on and four previously undocumented capabilities now cost their own characters. The win here is situation, not budget. I originally framed this as a budget reduction and that was wrong.

## Tests

`tests/test_mcp_surface_states_its_own_behaviour.py`, 10 tests, derived from what the code emits rather than restating strings, following `test_hosted_instruction_truth`'s reasoning about `binding`. Add a `FailureKind` or rename a context block and these fail rather than passing against stale text.

## Pre-existing failures, not from this PR

`uv run dev.py check` leaves 7 failures, all present on `main` at `f71f3c2` (verified by stash). All in `_sole_served_datasource` / `resolve_profile`: 3 in `test_admin_model.py`, 4 in `test_hosted_instruction_truth.py`. Worth a separate look, untouched here.

## Out of scope

Two Tier 1 items were deliberately skipped: the product capability statement and the rationale expansion on the verbatim-metric rule.


sandeep-agami:
Thanks — all four addressed in 9e0e92b. Three were real, and all three were this PR's own thesis turned back on it: a description asserting something the payload does not do.

**`list_datasources` `description` is conditional.** Confirmed: the served path emits it only via `**({"description": ...} if ds in descriptions else {})`, and the local path (`tool_list_datasources`, profile branch) never emits it at all — it builds `datasource` / `database_type` / `table_count` / `model_present` / `is_active`. "Each entry carries" was wrong on both paths. Now stated as conditional with the fallback named, because the cost is specific: routing without a schema call is this tool's entire purpose, so an agent that reads a missing key as an error re-adds the call it was meant to avoid.

**`default_filter` singular.** Correct, the key is `default_filters`. This is the same class as the `binding`/`bindings` defect `test_hosted_instruction_truth` already guards against — a backticked field name no payload carries sends a client looking for nothing. Reworded to "an entry in `default_filters`".

**Join item shape.** Correct and the strongest of the four. `assemble_receipt` seeds `{key: None for key in ("name", "cardinality", "confidence", "origin", "review_state", "signed_off_by", "signed_off_role", "signed_off_at", "cross_schema", "on")}` and spreads it into every join item, so all ten are always present and null unless status is `declared`. The declared shape stopped at four keys while the following sentences referenced `name` and `review_state`, and the shared instructions tell an agent to filter joins *on* `review_state`. So the surface named a field in its rule that its own shape said did not exist.

Now declared in full, including the present-but-null rule — that part is what makes "a join's `review_state` is null when its status is not 'declared'" read as a rule rather than a contradiction.

**Double-quote brittleness (both sites).** Fair. `ruff format --check` gates this repo so a single-quote refactor would not survive, but a test about payload shape that depends on the formatter to stay true is testing the formatter. Replaced with a quote-agnostic `_emits_key` helper.

Two tests added:

- `test_the_one_conditional_field_is_described_as_conditional`
- `test_the_join_item_shape_is_described_as_the_assembler_builds_it` — this one ast-parses `assemble_receipt`'s key tuple rather than copying it, verified to derive all 10 keys. Copying the list into the test would have created exactly the second place to drift this change exists to close.

12 tests in the file now, all passing. `uv run dev.py check` is clean apart from the 7 `resolve_profile` failures that pre-date this branch on `main` at `f71f3c2`.

PR #251, opened 2026-09-02
release: 0.7.0 — golden datasets

ashwin-agami:
Cuts **0.7.0** — 90 commits since `v0.6.9`, across nine PRs (#237, #239, #240, #241, #242, #243, #248, #249, #250).

## What this PR does

1. Bumps the version in the four places that must match: `.claude-plugin/marketplace.json` (×2), `plugins/agami/.claude-plugin/plugin.json`, `packages/agami-core/pyproject.toml`. The release tag must be `v0.7.0` or `release-pypi.yml`'s guard fails the publish.
2. Writes the `[0.7.0]` changelog section.

## The changelog was missing entirely

`[Unreleased]` was **empty**. All 90 commits landed without an entry, so a minor release had no notes at all — nothing told a user that `/agami-eval` and `/agami-save-golden` exist, or that a reconcile run now offers to keep the answers it proved. The section is written here rather than backfilled per-PR, since the nine PRs are one feature and read as one.

## Why minor, not patch

Purely additive, and verified as such rather than assumed:

- `git diff --name-status v0.6.9..HEAD` shows **no `D` and no `R` entries** — nothing deleted, nothing renamed.
- No argument, return shape, refusal rule, receipt key or CLI flag from 0.6.x moves.
- What is new: two skills (bringing the shipped count to nine), four library modules (`golden.py`, `comparator.py`, `golden_claims.py`, `golden_run.py`), five plugin scripts, and additive keys on `agami-reconcile`'s row record.

## Claims verified against the code, not the PR summaries

- **"The comparator executes nothing, opens no connection, cannot make a model call"** — its imports are stdlib + `sqlglot` + `ExecResult` as a type. No driver, no model client. Confirmed.
- **"Exit `2` means go look at the harness"** — `_CANNOT_START = 2` in `run_golden_eval.py`, and the verdict rests on `gating_failures`, never `failed`, so an unreviewed answer key cannot fail CI. Confirmed.
- **"The child's sandbox flags are asserted present by a test"** — `test_golden_run.py:577-579` and `612-614` assert `--tools ""`, `--strict-mcp-config` and `--setting-sources ""`. Confirmed. (My first check used `rg -rn`, where `-r` is `--replace` and silently rewrote the matches — worth noting since the mangled output looked like the flags were absent.)
- **"Cells canonicalize to a `(type_tag, value)` key"** — `comparator.py:171`. Confirmed.

## Verification

- `uv run dev.py check` — **5409 passed**, 12 skipped; ruff check + format clean; gitleaks clean
- `python -m build --sdist --wheel packages/agami-core` — builds `agami_core-0.7.0.tar.gz` + `.whl`
- `twine check` — both PASSED
- All four new modules confirmed present in the wheel
- `0.7.0` confirmed free on PyPI

## After merge

Merging publishes nothing. Both release workflows fire on `release: published`, so a `v0.7.0` release still has to be cut to reach PyPI and GHCR.


ashwin-agami:
`/code-review` found two documentation defects, both mine, both now fixed in `d8b0a93`.

**1. `--out` is not a flag on `run_golden_eval.py`.** The only `--out` in that file is a comment at line 723 noting that the renderer *beside* it takes one. A user following the release notes would have hit `unrecognized arguments: --out` and exit `2` — which in this tool's own vocabulary means *go look at the harness*, so the failure would have misled twice over.

The cause is worth recording: I built the list by grepping the file for flag-shaped strings, which swept up a sibling script's flags out of a comment. The parser is the only thing that defines the surface. Corrected against `add_argument` — `--profile`, `--dataset`, `--list`, `--tag`, `--rerun-failures`, `--top-k`, `--timeout-s` — which also restores `--top-k`, dropped for the same reason.

**2. `/agami-save-golden` was missing from the README.** The command table listed eight of the nine skills this release ships. By the changelog's own argument an empty dataset gates nothing, so the command that *fills* the dataset is the one that makes `/agami-eval` worth running, and it was absent from the front page. Added; the table now lists all nine.

Worth noting what this says about the release itself: the executable surface was clean, and both defects were in prose — which a release PR makes permanent, since the notes become an immutable GitHub release body once the tag is cut.

Gate re-run after the fixes: **5409 passed**, 12 skipped, ruff + gitleaks clean.

Issue #332, opened 2026-09-14
Scope checks match tables by bare name, so a same-named table in another schema passes as the declared one


ashwin-agami:
## What happens

With the semantic-model pass enabled (`AGAMI_GOVERNANCE_ENFORCED`), `check_table_scope` (`semantic_model/runtime.py`) decides whether a statement stays inside the model by comparing each referenced table's **bare name** (sqlglot `Table.name`, schema dropped), case-folded, against the model's declared table names. The schema a statement names is never compared.

So if the model declares `sales_data.orders` and the warehouse also has `staging.orders`, this passes the table-scope gate:

```sql
SELECT order_id FROM staging.orders
```

— and the receipt reports the table as a declared one. The same bare-name folding is used by the model's `bare_name` helper and `loader.build_table_index` (first-occurrence-wins on a name clash), so a name that exists in two schemas also resolves to whichever table was defined first when looked up.

## Why it matters

- **Governance accuracy, not the security boundary.** Per `SECURITY.md` the model pass is not the read-only gate and is off by default; the connecting role still bounds what can be read. But when an operator turns confinement on, they are relying on "only declared tables", and a same-named table in another schema defeats it silently, with a receipt that says otherwise.
- **Column scope** binds columns to declared tables by the same bare names and inherits the confusion.
- **Model lookups** can hand back the wrong table's columns, relationships and filters for a clashing name.

No known deployment has a clashing name today; this is latent until one does.

## Suggested direction

Make `(schema, name)` a table's identity wherever the model is compared or looked up, falling back to the bare name only where no schema is known:

- `check_table_scope` / column scope: compare the statement's `(db, name)` against declared `(schema_name, name)`; an unqualified reference matches a declared table only when its bare name is unique in the model (otherwise refuse as ambiguous — fail closed).
- `build_table_index` and friends: key on the qualified name; refuse or report a clash rather than first-occurrence-wins.
- Storage: `model_table` is keyed `(org_id, datasource, area, name)`, so two same-named tables in one area cannot both be stored — decide whether the key gains `schema`.

Related: #258 (serving the schema so generated SQL resolves).

## Tests to add

- A model declaring `sales_data.orders`; `SELECT … FROM staging.orders` is refused by table scope.
- An unqualified `FROM orders` with `orders` in two declared schemas is refused as ambiguous; with one, it passes.
- Two declared tables sharing a name in different schemas are both reachable by qualified name.

Issue #329, opened 2026-09-14
Let an organisation's administrator set the row cap and statement deadline


ashwin-agami:
## Today

The result row cap (`AGAMI_SQL_MAX_ROWS`, default 1000) and the per-statement deadline (`AGAMI_SQL_TIMEOUT_S`, default 30) are deployment-wide and read from the environment only. `_resolve_timeout_s` documents why that is deliberate: the fork path's child re-resolves from `os.environ`, so a request-scoped override the child cannot see would let the parent's supervisor bound sit below the budget the child enforces.

An administrator who wants a different limit for their organisation has to ask whoever runs the deployment to change an env var and redeploy.

## Ask

Let an organisation's administrator set both from the admin console, per organisation, without breaking the properties the env-only rule protects.

Points to decide rather than assume:

- **A deployment ceiling.** The limits protect a shared instance (one async worker serves every organisation). A per-org value must be clamped to an operator-set maximum — e.g. `AGAMI_SQL_MAX_ROWS_CEILING` / `AGAMI_SQL_TIMEOUT_S_CEILING` — so one tenant cannot stall the others.
- **One source for both sides of the fork.** Resolve the effective value once per call in the parent and hand it to the child explicitly (the `_pass_child_env` pattern), so the ordered bound family stays one budget.
- **The tool description.** Once the description states the limits (#326), it is built per organisation rather than once at start-up, and a change reaches new sessions only — say so in the console.
- **Audit.** Record who changed a limit and when, like other admin settings.
- **Storage.** Portable migration; absent means the deployment value.

## Out of scope

Changing the defaults; per-user limits.

ashwin-agami:
Done. The engine half shipped in agami-core 0.8.7: an embedder registers a per-organisation provider for the row cap and time limit, with the environment values as the default. The admin Settings screen that edits them is in the private product. As decided, there is no policy ceiling; values the engines cannot represent (a row cap of 2,147,483,647 or more, a time limit over seven days) are refused at save time.

Issue #258, opened 2026-09-04
Store table names fully qualified (schema.table) so generated SQL resolves and same-named tables in two schemas do not collide


ashwin-agami:
## What happens

Introspection captures each table's schema — `semantic_model/build.py:361` sets `schema=table.schema_name` — but the served model stores and returns the **bare** table name. `model_table` has no schema column at all (`area, datasource, doc, est_row_count, name, org_id`), so by the time `get_datasource_schema` answers, the qualifier is gone.

The caller then writes SQL using the name it was given, verbatim. On any warehouse whose tables are not in the connection's default schema, that statement cannot resolve:

```
Postgres execution error: relation "orders" does not exist
LINE 2: FROM orders
```

The table exists. It is `SALES_DATA.orders`, and nothing in the model or in the schema reply ever said so.

## Two separate consequences

**1. Generated SQL fails against any non-default schema.** Not intermittently — for every table in such a datasource, on every question. Observed on a Postgres datasource whose tables all sit in one named schema: the first statement of a turn fails, the caller retries with a qualified guess, and the turn costs an extra round trip. The recovery makes it look transient; it is systematic.

**2. Two tables with the same name in two schemas cannot both be represented.** `SALES_DATA.orders` and `staging.orders` reduce to one row keyed on `orders`. One silently replaces the other, or the write collides, depending on the path.

The second is already recognised elsewhere in this codebase. `model_store._relationship_key` documents "same-named tables in two schemas" as one of the collision cases that defeated an earlier key design, and its `endpoint()` helper deliberately renders `f"{schema}.{table}" if schema else f"{table}"`. So relationships already carry the qualified form for exactly this reason; tables do not.

## Expected vs actual

- **Expected:** a table's identity in the model is what identifies it in the warehouse. A name handed to a caller can be used in a statement without further qualification, and two tables that differ only by schema are two tables.
- **Actual:** the identity is the bare name. It is ambiguous within a datasource, and unusable in SQL wherever the schema is not the default.

## Suggested direction

Store and return the qualified name — `schema.table` where a schema is known, the bare name where one is not. That makes the model's name the warehouse's name, which is the property the rest of the system already assumes it has.

Points worth deciding rather than assuming:

- **Where the qualification lives.** A `schema` column on `model_table` keeps the two parts separable (useful for grouping and for display); folding it into `name` needs no migration but makes the field mean two things. `_relationship_key` already chose the folded form for its human-readable half, which is a precedent for one and not the other.
[...]

PR #340, opened 2026-09-14
release: 0.8.7

### LINKED ISSUE #327: A table_scope refusal on the wrong datasource says 'add the table to the model' when the table is declared in another datasource — name that datasource instead
## What happens

On a deployment with several datasources, the client sometimes sends SQL written against one datasource's tables to a different datasource. The table-scope gate refuses it — correctly — but the remediation it returns is wrong for this case:

> query references table(s) not in the semantic model: invoices, items — only tables declared in the model may be queried.
> **Add the table to the model (agami-connect / '/agami-model'), or remove it from the query.**

The tables *are* in a model — another datasource's, in the same org. The client cannot add tables to a model, so it guesses: on one question it sent the statement to the wrong datasource twice before landing on the right one (~31 s lost). Over two weeks on a three-datasource test deployment, 4 of 14 `execute_sql` failures were this.

## Why

1. **The client chooses blind.** `list_datasources` returns `description` precisely so the client can answer "which datasource does this question touch?" (`tools.py`, `tool_list_datasources`). On the affected deployment every datasource's description was empty, so the choice was made on names alone.
2. **The gate can't see past its own datasource.** `check_table_scope` (`semantic_model/runtime.py`) knows one datasource's model, so "not here" is all it can say, and the remediation falls back to "add it to the model".

## Suggested change

**A. Point the refusal at the right datasource (tool layer, not the gate).** In `tool_execute_sql`, after a `table_scope` refusal, look the refused bare table names up in the same org's other datasources' models:

- all found in exactly one other datasource → replace the remediation: *"`invoices`, `items` are declared in datasource `billing` — read its schema and run the query there."*
- split across datasources → say so, and that one statement can't join across datasources.
- not found anywhere → keep today's remediation.

This keeps the gate single-datasource by design, and discloses nothing new: `list_datasources` already enumerates every datasource in the org to any caller, and the refusal still echoes only the table names the caller wrote (the gate's "echo, never enumerate" property holds).

Do **not** re-route the statement automatically: that silently changes what the answer is computed from, and the client should read the target datasource's schema before its SQL can be trusted.

**B. Make an empty datasource description visible at model time.** Warn during model build/deploy when a datasource has no `description`, since that is the field the client uses to pick one.

## Related, small

These refusals are logged with an empty `tool_calls.datasource` — as were about half of all failed calls on a second deployment — which makes this kind of review harder. Record the datasource the call resolved to.

## Tests

- Refused tables all declared in one other datasource of the same org → remediation names that datasource.
- Tables split across two other datasources → remediation says they can't be joined in one statement.
- Table declared nowhere → remediation unchanged.
- A datasource in a *different* org declaring the table is never named.
- Deploying a model with an empty description emits the warning.

## Success check

On a multi-datasource deployment, a `table_scope` refusal is followed by a successful call on the named datasource on the next attempt, and repeated refusals of the same statement disappear.

ashwin-agami:
Cuts `0.8.7`. Version-only: bumps `marketplace.json` (2), `plugin.json`, `packages/agami-core/pyproject.toml`, and dates `[Unreleased]` in the CHANGELOG.

## What 0.8.7 releases
- **Schema in the summary tier, and `search_path` for a single-schema model** (#333, refs #258).
- **Datasource routing** (#336, closes #327): an omitted `datasource` is refused when several are served; a table-scope refusal names the datasource that declares the table; `sm set-description` and onboarding ask for a datasource description.
- **Per-organisation row cap and statement time limit** (#334, refs #329) and the public `tools.statement_limit_is_usable` (#338).

#335 (qualified table scope, #332) is not in this release; it follows in 0.8.8.

Publishing is a separate step: merging ships nothing until a `v0.8.7` GitHub release is cut.


ashwin-agami:
Superseded: cut again from main after #342 and #343 merged so the 0.8.7 notes are complete.

PR #351, opened 2026-09-15
CLAUDE.md: open PRs as drafts, base them on main, hold review scope

ashwin-agami:
Puts the PR workflow agreed today where every contributor and every Claude session reads it, so CI and review rounds are not spent on runs nobody merges:

- open PRs as drafts (CI skips tests on drafts since #350), handle Copilot on the draft in one push, then mark ready;
- base PRs on main, since CI rules come from the base branch;
- keep review rounds to the issue, fix same-shaped paths together, one follow-up issue for the rest;
- reply to each review comment;
- release PRs are version strings and the changelog heading only, and a PR needing a new release waits for PyPI.


ashwin-agami:
Replies to Copilot's suppressed comments, both fixed in 3b4ec8b:
- Drafts: now says CI skips the test *work* on drafts and the required checks still report, rather than skipping the jobs.
- Releases: now says a PR that *depends on* a new agami-core release (not the release PR itself) waits for PyPI.
- Same-shaped fix from agami#126's review: "the full suite runs once" is now "every applicable check runs once", since release-only PRs skip tests even when ready.

PR #378, opened 2026-09-17
ACE-152: MCP SDK 2 on the HTTP transport

vishalkalbi27:
Spec: ACE-152

## Summary

Moves the HTTP MCP transport from SDK 1.x (security fixes only) to `mcp>=2.2,<3`. Clients on either protocol version (2025-06-18 or 2026-07-28) see the same tools and answers, with two deliberate changes:
- **A crashed tool's reason no longer reaches the client.** The client gets `isError` `Error executing tool <name>`. The reason goes to one ERROR log record with its traceback, and to the new operator-only column `tool_calls.error_detail`.
- **A hidden or absent tool now answers JSON-RPC `-32602 Unknown tool: <name>`**, byte-identical for both on both protocol versions. This matches the stdio harness (REQ-002).

It also adds cache hints on `tools/list` and `server/discover`, and a host/origin allowlist on `/mcp`.

## Changes

- **`pyproject.toml`:** the server extra requires `mcp>=2.2,<3` and declares `jsonschema`.
- **`mcp_http.build_server`:**
  - **Handlers:** registered through SDK 2's constructor, returning explicit result objects.
  - **Outcome policy:** any exception escaping the call handler answers the fixed text, including a failed audit write. The call still fails (ACE-097).
  - **Argument validation:** SDK 2's low-level server stopped validating `arguments` against `inputSchema`, so the transport does it now, with 1.x's text, *after* the visibility check. A hidden tool with bad arguments no longer answers differently from an absent one. As before, a validation refusal writes no audit row.
  - **Validators:** built once at startup on an empty `referencing.Registry`. A consumer tool with an invalid schema fails at startup, and a remote `$ref` is never fetched. 1.x's `jsonschema.validate` would fetch one.
- **`mcp_http.create_app`:**
  - **Allowlist:** `TransportSecuritySettings` built from `PUBLIC_BASE_URL`: lowercased, default port optional, plus loopback names when the base URL is loopback.
  - **Cache hints:** `private` with a 60s TTL, on `tools/list` and `server/discover`.
- **Migration `027_tool_calls_error_detail.sql`:** adds `tool_calls.error_detail`. `contracts.ToolCallRecord`, `tools.record_tool_call` (bounded by `AUDIT_ERROR_DETAIL_MAX_CHARS`, only on raised calls) and `DbActivitySink` carry it. It is not added to `_TOOL_CALL_COLS`, so no page shows it.
- **Tests:**
  - **Snapshots:** taken on SDK 1 before the pin moved (`tests/fixtures/ace152/`).
  - **Transport suite:** a new suite runs the real app in-process on both protocol versions (`tests/mcp_eras.py`, `test_ace152_sdk2_transport.py`). Its leak scans read raw response bytes.
  - **Dependency test:** the extra's dependency closure contains no OpenTelemetry SDK or exporter.
- **CHANGELOG:** one entry under Unreleased.

### Existing tests changed, and why
- **`test_ace097_recording_mandatory::test_the_transport_no_longer_swallows_it_either`:** it asserted that the audit sink's error text reached the client. The spec decision forbids that. It now asserts the call still fails with the fixed text and that the reason is absent from the raw body.
- **`test_per_org_statement_limits.py` (3 tests) and `test_thread_id_required.py::_served_schema`:** SDK 2 dropped `server.request_handlers[...]`, so they use `get_request_handler("tools/list")`. Same assertions.
- **`base_url=` on every `TestClient` that POSTs `/mcp`:** needed now that the allowlist rejects `Host: testserver`. No assertion changed.
- **`test_tool_calls::test_a_raising_tool_is_still_logged_as_an_error`:** gains an `error_detail` assertion.
- **Comments:** those crediting "the MCP SDK" with argument validation now point at the transport.

### Findings worth knowing (not changed here)
- **Extra listing per call:** on 2026-07-28 the SDK runs `on_list_tools` before every `tools/call` that carries arguments, for its `Mcp-Param-*` header check. That adds a visibility-predicate pass per tool on each call, and there is no SDK switch to turn it off.
- **Browser clients:** one sending an `Origin` other than `PUBLIC_BASE_URL`'s now gets 403. claude.ai and other server-to-server clients send no Origin.
- **Reverse proxies:** a proxy must pass the public `Host` through. Caddy does by default; nginx's default `proxy_set_header Host $proxy_host` does not.
- **Legacy `tools/list` leak:** on 2025-06-18, an exception raised inside `tools/list` still reaches the client as `str(e)`, exactly as on 1.x. The outcome invariant covers tool calls only, so this goes in a follow-up issue, together with recording validation refusals.
- **Sensitive `error_detail` content:** it can contain driver text such as a host or user name, or a quoted data value. It stays operator-only, and any page that later shows it must redact it.
- **Byte-exact snapshots:** criterion 10 is byte-exact on 2026-07-28 but compared as parsed JSON on 2025-06-18. SDK 2 reorders the top-level `inputSchema` keys there; the values are unchanged. This is recorded in the spec Decisions.
- **Merge conflict with #377:** expect one in the `tool_calls` INSERT (migration 026). This PR uses 027.
- **Downstream consumers:** those whose tests POST `/mcp` through `TestClient` without `base_url=` will get 421 once they upgrade.

## Checklist
- [x] `Spec: ACE-152` referenced; diff within the spec's scope (no stdio, OAuth, or clarifying-question changes; no tool name, description or schema change)
- [x] Tests added or updated for every success criterion, on both protocol versions; no test weakened or deleted
- [x] `uv run dev.py check` green locally (6331 passed, 14 skipped; ruff and gitleaks clean); `uv run dev.py cover` 100% of changed lines
- [x] Postgres e2e (`tests/e2e`, including migration 027) passed against the compose Postgres
- [x] Live smoke test on both protocol versions: tools listed, cache hints on 2026-07-28 only, foreign Host 421, foreign Origin 403, `.well-known` unaffected
- [x] Review panel (correctness, security, Agami rubric) run; must-fixes addressed in `232f9ec`
- [x] Synthetic placeholders only; no secrets


vishalkalbi27:
Both suppressed comments handled in 6778c66.

**`mcp_http.py:767` — loopback aliases skipped on an https base URL.** Correct, and fixed. Reproduced with `PUBLIC_BASE_URL=https://localhost:5173`: `Host: localhost:5173` was served and `Host: 127.0.0.1:5173` got 421. `_is_loopback` is deliberately http-only, because it answers the "TLS required except on loopback" cookie rule, so `_transport_security` no longer asks it. It decides from the base URL's host, and the added origins now carry the base URL's own scheme instead of a hard-coded `http://`. `_is_loopback` itself is unchanged, since widening it would relax a different control. New tests cover an https loopback base serving `localhost`, `127.0.0.1` and `[::1]` while a foreign name still gets 421, and that the added origins are `https://`. The contract and the spec's Decisions record the change from "when `_is_loopback` holds" to "when the host is a loopback name".

**`mcp_http.py:838` — the authenticated seam caller.** The failure described doesn't occur: `test_a_refusing_resolver_gives_403_not_500` passes. Our auth middleware, which runs the org resolver, sits in front of the MCP mount, so the resolver's 403 (and the unauthenticated 401 in `test_build_app_still_serves_the_same_auth_challenge`) is returned before the host check is reached. Both tests now pass `base_url=BASE` anyway: it matches every other transport caller and keeps them from resting on middleware order. No test override is added — that would be a setting nobody sets in production; downstream in-process callers get the same one-line `base_url=` change when they adopt this release.

Full gate green after the fix: 6334 passed, 14 skipped, ruff and gitleaks clean, patch coverage 100%.

vishalkalbi27:
### Copilot's two suppressed findings: both already fixed, with tests

Re-checked on the current head; each was addressed after the 2026-09-17 review that raised it, so neither needs a change. Evidence rather than assertion:

**1. "`tests/test_tool_extension_seam.py:236` still creates `TestClient(...)` without `base_url=BASE`, so the host allowlist returns 421 before the resolver's 403."**

That client now passes `base_url=BASE` (`test_a_refusing_resolver_gives_403_not_500`), and the file is green: `tests/test_tool_extension_seam.py` — **18 passed**. The 403 assertion stands.

**2. "`_is_loopback()` is true only for `http://`, so `PUBLIC_BASE_URL=https://localhost:5173` never adds `127.0.0.1` and `[::1]`, and they get 421."**

`_transport_security` no longer asks `_is_loopback`: the aliases follow the **host**, and the origins take the **base's own scheme**, with the local-TLS case named in the comment there. Covered by `test_an_https_loopback_base_url_serves_every_loopback_name` and `test_an_https_loopback_base_url_allows_its_own_scheme_as_origin` — **11 passed** in that selection.

`_is_loopback` itself is unchanged on purpose: it answers a different question (whether plain http is acceptable for the admin cookie), which is still an http-only question.

